PR #70, opened 2021-09-26
Initial work on statistics page

brian-norman:
Currently doesn't compile because of the following error

```
[Vue warn]: Error in render: "TypeError: undefined is not an object (evaluating 'child.tag')"

found in

---> <Home> at src/views/Home.vue
       <App> at src/App.vue
         <Root>
```

Specifically fails to compile when I add `export default class StatisticsView extends Vue {`

Any ideas @cole-adams ?

### DIFF
--- a/src/views/StatisticsView.vue
+++ b/src/views/StatisticsView.vue
@@ -1,0 +1,51 @@
+<template>
+    <div>
+        <div class="title">
+            <router-link to="/"><mdicon class="back" name="arrow-left" size="40"/></router-link>
+            <h1>Cobresun Statistics</h1>
+        </div>
+
+        <div>
+            <h2>Most Loved Movie</h2>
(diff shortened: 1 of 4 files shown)


src/views/StatisticsView.vue
```diff
@@ -0,0 +1,51 @@
...
+            <h2>Most Loved Movie</h2>
+            <p>{{mostLovedMovie}}</p>
+        </div>
+    </div>
+</template>
+
+<script lang="ts">
+    import { Vue } from 'vue-property-decorator'
+    import axios from 'axios'
+
```
cole-adams, line 18: Right here you need to add a component decorator (meaning @Component({})) like in the other components

cole-adams (changes requested):

PR #78, opened 2022-05-21
Watchlist gallery

brian-norman:
# Need Help!
- [x] Poster heights are no longer all the same, looks weird
- [x] Button icons are not centered
- [x] The cards should appear on a new line below the randomize and add movie buttons

### DIFF
--- a/src/components/MoviePosterCard.vue
+++ b/src/components/MoviePosterCard.vue
@@ -1,0 +1,51 @@
+<template>
+  <div>
+    <div
+      v-if="highlighted"
+      class="border-4 border-highlightBackground rounded w-40 mb-4"
+    >
+      <div class="flex flex-col h-full">
+        <img :src="moviePosterUrl">
+        <div class="px-2 pb-2 flex flex-col h-auto flex-grow">
+          <div class="my-2 flex flex-grow items-center justify-center">
+            <h3
+              class="font-semibold h-min"
+              style="height: min-content"
+            >
+              {{ movieTitle }}
+            </h3>
+          </div>
+          <slot />
+        </div>
+      </div>
+    </div>
+
+    <div
+      v-else
+      class="border-2 border-gray-200 rounded w-40 mb-4"
+    >
+      <div class="flex flex-col h-full">
+        <img :src="moviePosterUrl">
+        <div class="px-2 pb-2 flex flex-col h-auto flex-grow">
+          <div class="my-2 flex flex-grow items-center justify-center">
+            <h3
+              class="font-semibold h-min"
+              style="height: min-content"
+            >
+              {{ movieTitle }}
+            </h3>
+          </div>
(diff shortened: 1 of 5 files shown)


src/components/MoviePosterCard.vue
```diff
@@ -0,0 +1,31 @@
...
+          </div>
+          <slot />
+        </div>
+      </div>
+    </div>
+  </div>
+</template>
+
+<script setup lang="ts">
+    const { movieTitle, moviePosterUrl } = defineProps<{
```
cole-adams, line 26: Generally don't indent the content of the script

src/views/WatchListView.vue
```diff
@@ -40,54 +40,47 @@
...
+        <div class="grid grid-cols-auto justify-items-center">
+          <MoviePosterCard
+            v-for="movie in watchList"
+            :key="movie.movieId"
+            :movie-title="movie.movieTitle"
+            :movie-poster-url="movie.poster_url"
+            :highlighted="movie.movieId === nextMovieId"
+          >
+            <div class="grid grid-cols-2 gap-2">
+              <v-btn @click="reviewMovie(movie.movieId)">
```
cole-adams, line 52: Adding 'flex justify-center' classes to these buttons will center the icons

src/components/MoviePosterCard.vue
```diff
@@ -0,0 +1,31 @@
+<template>
+  <div>
```
cole-adams, line 2: The height issue comes about since this div is not 100% in height. You can either add 'h-full' to this div or just remove it altogether and have its direct child be the root element of this component

src/views/WatchListView.vue
```diff
@@ -40,54 +40,47 @@
           <mdicon name="dice-multiple-outline" />
```
cole-adams, line 40: I'll be honest, I'm not sure how the alignment worked before. To fix, probably best to remove the 'float-left' from the buttons and surround them in their own div. Then the buttons and poster card grid should be on different lines. From there you can add whatever additional styling you want (justify etc.)

src/views/WatchListView.vue
```diff
@@ -97,42 +90,6 @@ axios
     nextMovie.value = response.data.nextMovie;
     nextMovieId.value = nextMovie.value.nextMovieId;
   });
-axios
```
cole-adams, line 100: This call was to fetch the members so we could display who added the movie to the watchlist. I'm not sure if you wanted to have that functionality in the new display, but I'm fine either way

cole-adams (changes requested):

src/views/WatchListView.vue
brian-norman: Getting rid of it since we weren't using it for now, can bring it back in the future if we need it (YAGNI)

PR #83, opened 2022-10-12
Cleaned up stats code and added custom chart

SunnyNagam:
hey


cole-adams (approved):

brian-norman:
Film revenue/budget is 0 for a bunch of movies, is there another field we can look at or should we filter those out?

brian-norman:
Can we use the same rounding as the reviews table/gallery? Maybe extract to fixed constant somewhere

brian-norman:
i wanna have histogram pls

PR #85, opened 2022-10-19
Convert modal to composition api

cole-adams:


brian-norman (approved):

brian-norman:
While your messing with modals could you make it so clicking outside the modal dismisses it 😅

PR #104, opened 2023-02-10
Stats page use detailed review and BOLD titles

SunnyNagam:
its like way faster this way 
158 requests -> 55 requests


brian-norman:
Seeing a bug where I click into the stats page, back out, and then click in again - blank screen

brian-norman (approved):

brian-norman (approved):

PR #113, opened 2023-03-14
Award feature

cole-adams:


brian-norman (approved):

netlify/functions/utils/tmdb.ts
```diff
@@ -2,66 +2,100 @@ import axios from "axios";
...
+      review.movieData = response.data;
+      review.movieData.poster_url = `${configuration.data.images.secure_base_url}w154${response.data.poster_path}`;
+    })
+  );
+  return reviews;
 }
 
-export async function getWatchlistItemMovieData(watchList: WatchListItem[]): Promise<WatchListItem[]> {
-    const configuration = await getTMDBConfig();
+export async function getDetailedMovie<T extends BaseMovie>(
```
brian-norman, line 66: It seems like there is potential to share code between `getDetailedMovie` and `getDetailedReviewData`

src/common/components/MovieSearchPrompt.vue
```diff
@@ -63,9 +69,14 @@ import { ref, computed, watch } from "vue";
...
 
-const { defaultList, defaultListTitle } = defineProps<{
+const {
+  defaultList,
+  defaultListTitle,
+  includeSearch = true,
+} = defineProps<{
   defaultList: MovieSearchIndex[];
   defaultListTitle: string;
+  includeSearch?: boolean;
```
brian-norman, line 79: I'm struggling to understand what `includeSearch` is and why it's needed

src/router/index.ts
```diff
@@ -56,6 +63,48 @@ const routes: Array<RouteRecordRaw> = [
...
     },
   },
+  {
+    path: "/club/:clubId/awards",
+    name: "Awards",
+    component: AwardsView,
+    meta: {
+      depth: 2,
+    },
+    children: [
```
brian-norman, line 73: Didn't know about this feature, should we be using it for all the pages within a club, and all the clubs within the whole site?

src/common/types/models.ts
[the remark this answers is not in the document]
brian-norman: The `services` are probably where I'm imagining transformations could occur. 

I'd love if that effort resulted in simpler netlify function code that simply acts as a data fetching layer from the DB, with no logic of it's own (since it's gotten pretty complex now)

netlify/functions/utils/tmdb.ts
cole-adams: Yes, I wrote this so it could be used in place of `getDetailedReviewData` out of the box, I just forgot to swap it out in the reviews endpoint. Updated. Now the reviews endpoint also uses this function.

src/common/components/MovieSearchPrompt.vue
cole-adams: By default, this component searches TMDB to provide search results. In the nominations page, we only want this modal to display our review movies for the year, so I added this prop to control whether there are search results from TMDB.

src/common/types/models.ts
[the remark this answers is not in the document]
cole-adams: I agree `models.ts` has gotten bloated and could use some refactoring. The main difference between our `DB` (or I wrote them as `base` in this new feature) models and our view models is the addition of the movie data. Since the transformation requires us to hit an external API, I don't necessarily agree that we should be doing the transformation in the service layer, because that would require the client to query TMDB for every movie, and depending on the user's connection, this could be **significantly** slower than the server sending all the requests and responding with one piece of data. I do think the transformation should still occur in the function. 

In my opinion, the backend is where we _should_ be doing the heavy lifting on logic needed for data fetching/transformation, with the client side focused on the logic needed for display.

In terms of cleaning up the models, I think it might make sense to split our `DB` models from our `view` models into separate files, with the `view` models extending from the base models. I agree that the client side should only use the `view` models.

src/router/index.ts
cole-adams: Yeah, we could convert the rest of the routes to use this approach. Would be a cleaner approach for sure. I have created #114 to track this change.

PR #116, opened 2023-03-16
Backend router

cole-adams:
I know I should be working on awards upgrades, but I couldn't in my right mind create more endpoints in the backend with the current routing structure so....
- Created a new `Router` class so our backend routing takes a similar approach to ExpressJS. This greatly reduces the boilerplate code we need for adding endpoints.
- Separated out awards endpoints into separate files

Its worth noting, even after splitting up the awards file (duplicating imports) AND adding an entirely new 130 line class, there is still a 50-line total reduction in code.

@wesley-tang is the most experienced with backend so would appreciate his thoughts on this new approach


brian-norman (approved):
Bless your soul

PR #119, opened 2023-04-02
Implement basic querying in search

brian-norman:
You can search for production companies now by typing `company:name` or title like `title:name` or genre like `genre:name` and string those together to make more interesting searches. For example, find the comedy marvel movies like `company:marvel genre:comedy`.


src/common/searchMovies.ts
```diff
@@ -0,0 +1,70 @@
...
+ * 
+ * Incluidng multiple filters seperated by spaces will implicitly do an AND search between them.
+ * 
+ * TODO: Add support for OR searches.
+ * TODO: Create a new vue component for the search bar that highlights filters different colors.
+ * TODO: Make the watchlist and backlog use DetailedMovie[] so they can use the same search function and bar.
+ * 
+ */
+export function filterMovies(reviews: DetailedMovie[], searchQuery: string): DetailedMovie[] {
+    if (!searchQuery) {
```
cole-adams, line 21: Just wondering if this check is needed? By the type definition, it should never be undefined. If we're allowing it to be undefined, we should probably change the type definition to reflect that

src/common/searchMovies.ts
```diff
@@ -0,0 +1,70 @@
...
+      ? searchQuery.split(' ').reduce((acc, filter) => {
+          const [key, value] = filter.split(':');
+          acc[key] = value;
+          return acc;
+        }, {} as Record<string, string>)
+      : {};
+
+    // If there are filters, remove the filter and the value from the search query.
+    if (Object.keys(filters).length) {
+      searchQuery = searchQuery.replace(/(\w+:\w+\s?)/g, '');
```
cole-adams, line 38: Same with here, we should create a new variable we can manipulate inside the function body to avoid reassigning function parameters.

cole-adams:
I'm excited for this feature. This looks sick!

PR #130, opened 2023-04-18
Make search look better and use slash hotkey

brian-norman:
Its more like GitHub now hehe


cole-adams:
Love this feature! It looks like the text goes under the hotkey when its too long. Wondering if the hotkey icon should disappear on focus like in GitHub? 
Screen Shot 2023-04-18 at 4 30 33 PM

brian-norman:
> Love this feature! It looks like the text goes under the hotkey when its too long. Wondering if the hotkey icon should disappear on focus like in GitHub?
> Screen Shot 2023-04-18 at 4 30 33 PM

✅

Works great for when inputing the text, but this doesn't really solve for the slash coming back after unfocusing ... I'm not too worried about it tho, I think I want to eventually make it more like Github's modal anyways

cole-adams (approved):

Issue #107, opened 2023-02-14
Browser cache of reviews watchlist/backlog 


brian-norman:
We fire off an async request for updates but initially load the cached value

Then update view

PR #134, opened 2023-05-19
#107 Add caching to query client

cole-adams:
#107 
Wow it was so each, just needed to add more libraries


brian-norman (approved):
omg you did it 😱

PR #138, opened 2023-06-25
Reuse the ReviewView search bar in the watch list

brian-norman:
We can now search genre and company in the watch list / backlog


cole-adams (approved):
Looks good! Might be good to extract the search bar into its own component in the future to avoid duplicating the mounted and key press logic etc.

PR #141, opened 2023-09-24
Deep copy reviews list when making filteredReviews

brian-norman:
Fixes the watch list randomizing function by avoiding returning the original array


cole-adams (approved):
LGTM!

PR #148, opened 2023-11-26
Profile page

cole-adams:
* Added profile page accessible by clicking avatar in the nav bar
* Added ability to upload profile pictures


netlify/functions/member.ts
```diff
@@ -1,38 +1,99 @@
...
-      q.Map(
-        q.Paginate(q.Match(q.Index("members_by_email"), email)),
-        q.Lambda("X", q.Get(q.Var("X")))
-      )
+    const user = await faunaClient.query<Document<Member>>(
+      q.Get(q.Match(q.Index("members_by_email"), email))
     );
 
-    const responseBody = req.data[0].data;
+    // Upload to Cloudinary using the SDK
```
brian-norman, line 51: Can we pull the cloudinary stuff outside of this file? Would be nice to have some clean interface for replacing the users photo that this api call makes, so that in the future if we replace cloudinary with another provider we can just replace it there, and the api call doesn’t need to know why

src/common/types/club.ts
```diff
@@ -6,6 +6,7 @@ export interface Member {
   email: string;
   name: string;
   image: string;
+  assetId?: string;
```
brian-norman, line 9: What’s assetId

brian-norman (approved):
Epic

netlify/functions/member.ts
cole-adams: That's a good point. I've created an `imageRepository` file with the functions to perform the upload/delete so we should be able to hot swap out that if needed in the future.

As for file limits, the file limit is 6MB and that comes from the payload limit on the lambda function. I've added a warning on the frontend if the user tries to upload a large file. I don't think much can be done on the backend however. The lambda just crashes if it receives a request that is too large.

src/common/types/club.ts
cole-adams: `assetId` is the id of the user's profile image. I am storing it so that we can delete the old profile picture when the user uploads a new one.

brian-norman: Can we rename it so it’s clear it’s for the user’s image?

cole-adams: Sure. I've updated to `imageId` to match the `image` field in the user object

brian-norman:
LGTM

PR #164, opened 2024-03-13
Migrate backlog to postgres

cole-adams:
Changes:
* Added work, work_list, and work_list_item tables to support lists
* Migrated backlog items to a more generic list data type
* Remove outdated API docs (personally don't think we need the docs IMO)


migrations/data/20240303_MigrateBacklog.ts
```diff
@@ -0,0 +1,81 @@
...
+      .select("id")
+      .where("legacy_id", "=", club.data.clubId)
+      .execute();
+
+    // Ensure every club has a work list, even if empty
+    const [workListId] = await db
+      .insertInto("work_list")
+      .values({
+        club_id: cockroachClub.id,
+        type: "backlog", // Assuming type to distinguish this list
```
brian-norman, line 32: enum

cole-adams: Fixed

brian-norman (approved):
🚀

Issue #169, opened 2024-05-06
Database transcactions


cole-adams:
There has been a couple weird behaviours pop up where the list service attempts to insert a work that doesn't exist. I suspect this is due to a race condition where the work is getting deleted in between the service fetching the id and inserting into the work_list_item table.

To properly fix this, we should adjust our repository architecture to support database transactions. As an interim solution, on the frontend we will avoid removing from and adding to a different list at the same time.

PR #170, opened 2024-05-06
Improved loading UI for watchlist

cole-adams:
Changes:
* Add delete button to watchlist items
* Move delete button for backlog items for consistency
* Add loading overlay to movies that are added optimistically so further actions aren't performed
* Ensure delete/add functions are performed sequentially to avoid race condition on the backend (correct solution would be database transactions. I've logged #169 to be actioned in the future)
* Ensure lists can only be edited by users in club (whoops, I can't believe I forgot to include this in the lists PR)


wesley-tang (approved):
Fantastic, very fast, very cool

PR #171, opened 2024-05-22
Add time_added timestamp column to work_list_item and migrate dates to new column

cole-adams:
Made an oopsie when creating the date column for work_list_items. Should be a timestamp instead of a date. This PR makes the created_date optional (for future removal) and add the required (and generated) time_added column.


SunnyNagam (approved):
lgtm

PR #175, opened 2024-06-21
Add support for edit and delete reviews

cole-adams:
Changes: 
* Add tanstack table package (I'm sure much to @brian-norman's enthusiasm, this is step 1 in phasing out our own table class)
* Convert reviews page to use tanstack table instead of our home brewed concoction
* Added prettier as a direct dependency and installed opinionated Tailwind class ordering
* Moved review endpoints to a different file (still ironing out the separation of concerns between list items and reviews, this felt right in the moment)
* Adjusted type for reviews to return review id and created date rather than just score
* Added support for editing reviews by clicking on table entry
* Added support for deleting reviews by clicking on delete button from gallery view
(Full feature parity between views is coming soon)
* Cried a little when I looked at converting table in statistics page to tanstack table


brian-norman:
Test failed 😅

brian-norman:
I get this error when trying to run locally:

```
✘ [ERROR] [plugin dynamic-imports] Private name #props is not defined. (2:47)

    node_modules/kysely/dist/esm/index.js:85:14:
      85 │ export * from './migration/file-migration-provider.js';
         ╵               ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

  This error came from the "onDynamicImport" callback registered here

    ../../.nvm/versions/node/v19.9.0/lib/node_modules/netlify-cli/node_modules/@netlify/esbuild/lib/main.js:853:22:
      853 │         let promise = setup({
          ╵                       ^
```

cole-adams:
Hmm, I got the test passing but I'm not sure about your error. The deploy worked. Maybe try updating your netlify cli?

brian-norman:
Bug: I add a new review on the Reviews Page, click into the edit score field, and it's always pre-populating that field with the score of the movie we saw most recently. If I click down to edit any other movie now, it pre-populates the score with the score of the movie at the previous index. Smells like a classic off by one error after all the indexes are invalidated when we add a new movie 😛

brian-norman:
Bug: If I add a new movie on the Reviews Page, I am unable to switch between gallery view and table view. Also, it seems that if I add a new movie on the gallery view, it doesn't show up in the list

brian-norman (changes requested):
Mostly epic, some bugs reported

cole-adams:
Issues found should be working now

brian-norman (approved):

PR #182, opened 2024-12-17
Add caching to tmdb movie data

SunnyNagam:
Cache the movie details data from TMDB to a separate table so we don't need to make each movie call on page load.
image


netlify/functions/club/list.ts
cole-adams, line 1: Hey it doesn't seem like we're updating the movie details when we get a new work. We need to do that now we're storing all that information in the database

src/features/statistics/StatisticsView.vue
```diff
@@ -314,6 +433,7 @@ const calculateStatistics = () => {
 const setReviews = (isLoading: boolean) => {
   if (isLoading) return;
   movieData.value = fetchMovieData(reviews.value ?? []);
+  console.log(movieData.value);
```
cole-adams, line 436: Stray console log

cole-adams (approved):

cole-adams:
Don't forget to run the data migration after merging to prod. The schema migration gets run automatically

PR #183, opened 2024-12-22
Filter reviews by year

brian-norman:
In preparation for awards season, we can now filter reviews by the year we watched them! Just type `year:2022` or whatever to see it in action 😄 

Screenshot 2024-12-22 at 10 23 33 AM

### DIFF
--- a/src/common/searchMovies.ts
+++ b/src/common/searchMovies.ts
@@ -13,12 +13,6 @@ import { DetailedWorkListItem } from "./types/lists";
  *
  * Incluidng multiple filters seperated by spaces will implicitly do an AND search between them.
  * 
- * Supported filters:
- * genre: filters by genre
- * title: filters by title
- * company: filters by production company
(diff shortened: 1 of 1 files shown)


src/common/searchMovies.ts
```diff
@@ -5,12 +5,19 @@ import { DetailedWorkListItem } from "./types/lists";
...
  * @returns reviews filtered by searchQuery.
+ * 
  *
  * You can apply filters on the searchQuery with text:value. For example, to filter by title and genre, you can use:
  *
  * "title:jaws genre:horror"
  *
  * Incluidng multiple filters seperated by spaces will implicitly do an AND search between them.
+ * 
+ * Supported filters:
```
cole-adams, line 16: I personally don't like keeping documentation like this in the code, its too easy for this to get out of date. Do we need it? I think the code is readable enough to understand these are the available types?

PR #189, opened 2025-02-04
add share review page and restrict club access to members

SunnyNagam:
Share button appears on hover of the review (where garbage can would go if we were in edit mode):
image

image
image
image

And going to a club reviews page or something if you're not in the club now will redirect you to home.


netlify/functions/club/reviews.ts
```diff
@@ -64,4 +64,66 @@ router.put(
   },
 );
 
+router.get("/:workId/shared", async ({ clubId, params }, res) => {
```
cole-adams, line 67: Should we do a validation token similar to the join? As it sits, if I knew the work id I could access anyones review. Super low stakes though, could tackle in the future

netlify/functions/club/reviews.ts
```diff
@@ -64,4 +64,66 @@ router.put(
...
+
+  const work = {
+    id: workDetails.id,
+    title: workDetails.title,
+    type: workDetails.type,
+    imageUrl: workDetails.image_url ?? undefined,
+    externalId: workDetails.external_id ?? undefined,
+    externalData: hasValue(workDetails.overview)
+      ? {
+          adult: workDetails.adult,
```
cole-adams, line 91: Can we make this a util? overviewToExternalData or something? Just don't want to forget updating here if we update a field

netlify/functions/repositories/ReviewRepository.ts
```diff
@@ -123,6 +123,88 @@ class ReviewRepository {
...
+  async getReviewsByWorkId(clubId: string, workId: string) {
+    return db
+      .selectFrom("work_list")
+      .where("work_list.club_id", "=", clubId)
+      .where("work_list.type", "=", WorkListType.reviews)
+      .innerJoin("work_list_item", "work_list_item.list_id", "work_list.id")
+      .innerJoin("work", "work.id", "work_list_item.work_id")
+      .where("work.id", "=", workId)
+      .leftJoin("review", "review.work_id", "work.id")
+      .leftJoin(
```
cole-adams, line 136: Do we need all these joins for the work details? Does the endpoint get these in a separate call to the database?

src/features/reviews/components/TableView.vue
```diff
@@ -37,7 +37,7 @@
       <tr
         v-for="row in reviewTable.getRowModel().rows"
         :key="row.id"
-        class="h-20 bg-lowBackground"
+        class="group h-20 bg-lowBackground"
```
cole-adams, line 40: Whats this class doing?

src/features/reviews/views/ReviewView.vue
```diff
@@ -205,7 +208,24 @@ const columns = computed(() => [
...
+              class:
+                "opacity-0 group-hover:opacity-100 transition-opacity duration-100",
+            },
+            [
+              h(mdicon, {
+                name: "share",
+                class: "cursor-pointer hover:text-primary transition-colors",
+                onClick: () => {
+                  const url = `${window.location.origin}/share/club/${clubId}/review/${row.original.id}`;
+                  void navigator.clipboard.writeText(url);
```
cole-adams, line 223: I'd like to avoid the "void" promise assertion. In this case I would add a .then and a .catch statement and only show the toast success in the then, and a toast error in the error case

src/router/index.ts
```diff
@@ -19,6 +26,44 @@ import YearView from "@/features/awards/views/YearView.vue";
...
+const checkClubAccess = async (
+  to: RouteLocationNormalized,
+  from: RouteLocationNormalized,
+  next: NavigationGuardNext,
+) => {
+  const auth = useAuthStore();
+
+  // Wait for auth initialization to complete
+  while (auth.authLoading) {
+    console.log("waiting for auth to load");
```
cole-adams, line 38: Do we still need this logging?

netlify/functions/club/reviews.ts
cole-adams, line 1: Not sure if we want to do this in this PR or not, but to actually restrict clubs on the backend, we need to add the `secured` middleware to everywhere we're getting club data

src/router/index.ts
```diff
@@ -19,6 +26,44 @@ import YearView from "@/features/awards/views/YearView.vue";
...
+
+  // If not logged in, redirect to clubs
+  if (!auth.isLoggedIn) {
+    console.log("redirecting to clubs");
+    return next({ name: "Clubs" });
+  }
+
+  const clubId = to.params.clubId;
+  try {
+    const response = await auth.request.get<ClubPreview[]>("/api/member/clubs");
```
cole-adams, line 50: Does this request belowing in the auth store or somewhere else? Are we making this request at every nav?

src/stores/auth.ts
```diff
@@ -19,6 +20,9 @@ export const useAuthStore = defineStore("auth", () => {
       user.value = netlifyIdentity.currentUser();
       return token;
     },
+    onSettled: () => {
```
cole-adams, line 23: `useQuery` returns a ref `isLoading` so we shouldn't need to manage it on our own

cole-adams (changes requested):

netlify/functions/club/reviews.ts
SunnyNagam: True good idea or we could make the url use list item id or something so it's not consistent across clubs. idk i'll add it to the backlog

src/features/reviews/components/TableView.vue
SunnyNagam: enables the hover movie tooltip visibility

src/stores/auth.ts
SunnyNagam: I cannot make use of isLoading() since that will return false before the query is run for the first time on fresh page load, triggering router/index.ts where this is used to stop waiting for login and check club access. I need to track when it has finished successfully loading for the first time and this is the best way I found to do that in my testing but let me know if theres a better way

SunnyNagam: nvm found a better way lol, useQuery returns a ref `isInitialLoading` which is exactly what i was looking for

PR #190, opened 2025-02-04
secure join link away from clubid

SunnyNagam:
image
image
image

- creates new table club_invite with {token, club_id, expires_at}
- when opening settings page -> creates new token link that expires in 24hrs (replaces existing expired token or updates un-expired key if applicable)
- new join link links token in url with club if not expired

### DIFF
--- a/lib/types/generated/db.ts
+++ b/lib/types/generated/db.ts
@@ -11,11 +11,16 @@ export enum WorkType {
   movie = "movie",
 }
 
-export type Generated<T> = T extends ColumnType<infer S, infer I, infer U>
-  ? ColumnType<S, I | undefined, U>
-  : ColumnType<T, T | undefined, T>;
+export type Generated<T> =
+  T extends ColumnType<infer S, infer I, infer U>
+    ? ColumnType<S, I | undefined, U>
+    : ColumnType<T, T | undefined, T>;
 
-export type Int8 = ColumnType<string, bigint | number | string, bigint | number | string>;
+export type Int8 = ColumnType<
+  string,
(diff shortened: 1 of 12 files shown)


lib/types/generated/db.ts
cole-adams, line 1: Hey I don't see the table migration script in this PR? Is it missing?

netlify/functions/repositories/UserRepository.ts
```diff
@@ -89,6 +90,44 @@ class UserRepository {
       .where("user_id", "=", userId)
       .execute();
   }
+
+  async createClubInvite(clubId: string) {
```
cole-adams, line 94: I don't think this belongs in the user repository. Maybe move to club repsitory?

cole-adams (changes requested):

lib/types/generated/db.ts
SunnyNagam: Not really cause it's simply:
```
CREATE TABLE club_invite (
    token TEXT PRIMARY KEY,
    club_id BIGINT NOT NULL REFERENCES club(id),
    expires_at TIMESTAMP NOT NULL
);
```
i just run it on cockroachdb, i can make a migration for it too if it's recordkeeping thing

cole-adams: Yes please make a migration. If we ever need to spin up a new environment, we need to get to the current schema

cole-adams: The migration will run automatically on prod after merging

cole-adams (approved):

PR #193, opened 2025-03-18
Gallery Drawer and Table UI improvements

SunnyNagam:
image

the drawer opening animation on desktop is kind of a work in progress, its still a lil janky but idk how much more time i can burn on trying to make this smoother

image

also the tooltip for movie details now is only enabled on desktop and not mobile


cole-adams (commented):
Looks good, need to merge other PR first

PR #206, opened 2025-12-04
Add expandable movie description in mobile review details

brian-norman:
- Movie descriptions now show only 2 lines initially with a "Read more" button
- Automatically detects if description exceeds 2 lines using height measurement
- "Read more" button only appears when description is longer than 2 lines
- Clicking "Read more" expands to show full description
- "Show less" button appears when expanded to collapse back to 2 lines
- Uses Tailwind's line-clamp-2 utility for clean text truncation
- Resets expansion state when switching between movies


src/features/reviews/components/MovieDetailsDrawer.vue
```diff
@@ -249,6 +279,47 @@ watch(
...
+  () => {
+    isDescriptionExpanded.value = false;
+    void checkDescriptionHeight();
+  },
+  { immediate: true },
+);
+
+// Also check when drawer opens
+watch(
+  () => props.isOpen,
```
cole-adams: `isOpen` doesn't affect the layout. We don't need this watch if we key the component to change on a new movie. We should just check the height in an `onMounted` hook

PR #208, opened 2025-12-04
Fix bottom sheet scroll issue on mobile

brian-norman:
Prevent background content from scrolling when the movie details bottom sheet is open on mobile by:
- Locking body scroll (overflow: hidden) when drawer opens on mobile
- Restoring body scroll when drawer closes
- Handling screen size changes to adjust scroll lock appropriately
- Ensuring cleanup on component unmount

The fix only applies to mobile view (<768px) to avoid affecting the desktop drawer behavior.


src/features/reviews/components/MovieDetailsDrawer.vue
```diff
@@ -235,20 +235,41 @@ onMounted(() => {
...
+      }
     }
   },
 );
 
+// Handle body scroll lock when screen size changes
+watch(isMediumScreen, (newValue) => {
+  // If drawer is open and we resize to desktop, unlock body scroll
+  if (newValue && props.isOpen) {
+    document.body.style.overflow = "";
```
cole-adams: Would like to see this moved into a util if we build more bottom sheets

cole-adams (approved):

PR #221, opened 2026-01-05
feat: Configure better auth to support login with google

cole-adams:


brian-norman (approved):
I think your migration broke prod lol

PR #229, opened 2026-01-15
Migrate from netlify functions 3 to 5

brian-norman:
```markdown
⏺ Based on my analysis of your codebase and the v5 upgrade, here are the opportunities available:

  Current State

  Your project uses the legacy Netlify Functions API with:
  - Handler, HandlerEvent, HandlerContext types (AWS Lambda-style)
  - Custom Router system for path parameters and middleware chaining
  - Manual response object construction

  Interestingly, your auth.ts already converts between the legacy API and modern Web standards (Request/Response) for BetterAuth compatibility (netlify/functions/auth.ts:15-24).

  Benefits Already Gained from v5

  ✅ Massive dependency reduction: v5 reduced package size from 82 MB to 84 KB (~1000x smaller)
  ✅ Fewer transitive dependencies: From 311 to just 1

  New Features Available in v5

  1. Modern Functions API (Web Standards)

  Instead of AWS Lambda events, use standard Request and Response:

  import type { Context, Config } from "@netlify/functions";

  export default async (req: Request, context: Context) => {
    const body = await req.json();
    return new Response(JSON.stringify({ success: true }), {
      status: 200,
      headers: { "Content-Type": "application/json" }
    });
  };

  Benefits:
  - Web platform standards (same as fetch API)
  - Better TypeScript types
  - Same API as Netlify Edge Functions
  - No custom conversions needed (unlike your current auth.ts approach)

  2. Built-in Config Export for Routing

  Replace your custom router with native configuration:

  export const config: Config = {
    path: ["/api/club/:clubId", "/api/club/:clubId/*"],
    method: ["GET", "POST"]
  };

  This provides:
  - Path parameters natively
  - HTTP method filtering
  - Multiple path patterns per function
  - URLPattern syntax support

  3. Zero-Config Streaming

  Return ReadableStream directly for progressive responses:

  export default async () => {
    const stream = new ReadableStream({
      start(controller) {
        controller.enqueue(new TextEncoder().encode("data"));
        controller.close();
      }
    });
    return new Response(stream);
  };

  4. Background Functions

  Add -background suffix for long-running tasks (immediately returns 202):

  // netlify/functions/process-background.mts
  export default async (req: Request, context: Context) => {
    await longRunningTask(); // Continues after response sent
  };

  5. Better Environment Variables

  Access via Netlify.env.get() with better type safety:

  const dbUrl = Netlify.env.get("DATABASE_URL");

  Opportunities for Your Project

  Low-Effort Improvements

  1. Add streaming to any endpoints that could benefit from progressive rendering
  2. Use background functions for async tasks like email sending or image processing
  3. Upgrade netlify-cli globally to v23.13.3 (you're on v23.12.3)

  Medium-Effort Migration

  Migrate to modern API - This would provide significant benefits but requires substantial refactoring:

  Pros:
  - Eliminate conversion logic (see your auth.ts:15-24)
  - Use native routing features instead of custom Router
  - Better DX with web standards
  - Same API for functions and edge functions
  - Improved TypeScript inference

  Cons:
  - Must rewrite all function handlers
  - Custom router middleware would need reimplementation
  - Path parameter extraction works differently
  - Your sophisticated middleware chaining (validClubId, secured, loggedIn) needs adaptation

  Migration path: Netlify supports both APIs simultaneously, so you could migrate incrementally:
  1. New functions use modern API
  2. Gradually convert existing functions
  3. Keep legacy functions until fully migrated

  Recommendation

  Given your sophisticated routing system with middleware chaining, I'd recommend:

  Short term:
  - Keep the legacy API (it's still fully supported)
  - You already got the main v5 benefit (smaller package size)
  - Consider streaming for specific use cases

  Long term:
  - Plan migration to modern API as technical debt cleanup
  - Start with simpler functions (like auth.ts which already uses Request/Response internally)
  - Evaluate whether native Config routing can replace your custom Router, or if you need to maintain some custom logic

  The legacy API isn't going anywhere immediately - Netlify's deprecation timeline extends through 2025+ with ample warning. Your current implementation is solid and functional.

  Sources:
  - https://github.com/netlify/primitives/releases/tag/functions-v5.0.0
  - https://developers.netlify.com/guides/migrating-to-the-modern-netlify-functions/
  - https://docs.netlify.com/build/functions/get-started/
  - https://www.npmjs.com/package/@netlify/functions
  - https://www.netlify.com/blog/introducing-netlify-functions-2-0/
  ```

### DIFF
--- a/src/common/components/DeleteConfirmationModal.vue
+++ b/src/common/components/DeleteConfirmationModal.vue
@@ -1,0 +1,43 @@
+<template>
+  <Teleport to="body">
+    <v-modal v-if="show" size="sm" z-index="60" @close="emit('cancel')">
+      <div class="flex flex-col gap-4">
+        <h2 class="text-xl font-bold">{{ title }}</h2>
+        <p>{{ message }}</p>
+        <div class="flex gap-3">
+          <button
+            class="flex-1 rounded-md bg-gray-600 py-3 font-bold text-white hover:brightness-110"
+            @click="emit('cancel')"
+          >
+            Cancel
+          </button>
+          <button
+            class="flex-1 rounded-md bg-red-500 py-3 font-bold text-white hover:brightness-110"
+            @click="emit('confirm')"
+          >
+            Delete
+          </button>
+        </div>
+      </div>
+    </v-modal>
+  </Teleport>
+</template>
+
+<script setup lang="ts">
+withDefaults(
+  defineProps<{
+    show: boolean;
+    title?: string;
+    message?: string;
+  }>(),
+  {
+    title: "Delete Review",
+    message: "Are you sure you want to delete this review? This action cannot be undone.",
+  },
+);
+
+const emit = defineEmits<{
+  (e: "confirm"): void;
+  (e: "cancel"): void;
+}>();
+</script>
--- a/src/common/components/VBackdrop.vue
+++ b/src/common/components/VBackdrop.vue
@@ -13,7 +13,7 @@
 <script setup lang="ts">
 const props = withDefaults(
   defineProps<{
-    zIndex?: "40" | "50";
+    zIndex?: "40" | "50" | "60";
   }>(),
   {
     zIndex: "50",
@@ -24,7 +24,10 @@ const emit = defineEmits<{
   (e: "close"): void;
 }>();
 
-const zIndexClass = props.zIndex === "40" ? "z-40" : "z-50";
+const zIndexClass =
+  props.zIndex === "40" ? "z-40" :
+  props.zIndex === "60" ? "z-[60]" :
+  "z-50";
 
 const handleClose = () => {
   emit("close");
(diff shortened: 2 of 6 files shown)

### CODE REVIEW

**Comment by brian-norman** [comment]
@cole-adams I was running into my usage limit so here is a plan for migrating the rest lol: https://github.com/Cobresun/movie-club/issues/230

File: netlify/functions/utils/auth.ts [hunk 1]
```diff
@@ -54,11 +73,51 @@ export const auth = betterAuth({
...
+    // Always trust localhost for development
+    if (requestOrigin.startsWith("http://localhost:")) {
+      if (!trusted.includes(requestOrigin)) {
+        trusted.push(requestOrigin);
+      }
+      return trusted;
+    }
+
+    // Extract Netlify site identifier (if URL is a Netlify URL)
+    const siteIdentifier = extractNetlifySiteFromUrl(process.env.URL);
```
cole-adams: This url is not a netlify URL, this won't work

cole-adams: You can hardcode the deploy preview patter with our site id. Maybe we should have the site id as an envariable? We should also see if we can find a way to avoid it on production for security reasons

cole-adams (changes requested):

PR #255, opened 2026-01-29
Add support for reordering watchlist and backlog

brian-norman:


src/features/watch-list/components/ClubBacklog.vue
```diff
@@ -132,6 +141,20 @@ const sortedBacklog = computed(() => {
   ];
 });
 
+const draggableList = ref<DetailedWorkListItem[]>([]);
+watch(
+  sortedBacklog,
```
cole-adams: The `sortedBacklog` property puts the randomly selected item at the start. If you then performed a drag, it would complete re-order this list to have the random movie first. Would this be expected? Seems a little weird to me. Maybe we don't allow dragging after a random pull? Or something else?

src/features/watch-list/components/WatchList.vue
```diff
@@ -167,6 +174,20 @@ const sortedWatchList = computed(() => {
   return displayWatchlist.value;
 });
 
+const draggableList = ref<DetailedWorkListItem[]>([]);
+watch(
+  sortedWatchList,
```
cole-adams: The sorted watchlist puts the next movie first. If you drag something to the first slot, it quickly moves it to the second spot. Seems confusing to me. Is there a better way to do this UX

src/service/useList.ts
```diff
@@ -93,6 +93,44 @@ export function useDeleteListItem(clubId: string, type: WorkListType) {
...
+        type,
+      ]);
+      queryClient.setQueryData<DetailedWorkListItem[]>(
+        ["list", clubId, type],
+        (currentList) => {
+          if (!currentList) return currentList;
+          const itemMap = new Map(currentList.map((item) => [item.id, item]));
+          return workIds
+            .map((id) => itemMap.get(id))
+            .filter((item): item is DetailedWorkListItem => !!item);
```
cole-adams: minor: can use `.filter(isDefined)`

src/features/watch-list/components/ClubBacklog.vue
```diff
@@ -119,24 +142,34 @@ const closePrompt = () => {
...
+};
 
 const selectRandom = () => {
-  if (!backlog.value) return;
+  if (!backlog.value || backlog.value.length === 0) return;
   clearSearch();
-  const selectedIndex = Math.floor(Math.random() * backlog.value?.length);
-  const randomMovie = backlog.value[selectedIndex];
-  selectedMovie.value = randomMovie;
+  if (backlog.value.length === 1) {
```
cole-adams: This behaviour seems a little weird to me, that we wouldn't pop up the modal at all if there's only one. A bit of a dumb case, but I would want all the situations when you click the random button to be the same

PR #295, opened 2026-02-21
Fix iOS keyboard Done button not submitting score

brian-norman:
On iOS, tapping "Done" on the keyboard fires a blur event rather than a
keypress event, so the score was silently discarded. Submit the score on
blur to handle iOS Done, and use a submittedViaKeyboard flag to prevent
double-submission when Enter is used on desktop (which also triggers blur
when the input is removed from the DOM).

https://claude.ai/code/session_01491QMhaTeqmJMweetWmiLz


src/features/reviews/components/ReviewScore.vue
```diff
@@ -79,4 +79,25 @@ const submitScore = (score: number) => {
...
+const handleKeyEnter = () => {
+  submittedViaKeyboard.value = true;
+  submitScore(parseFloat(scoreModel.value));
+};
+
+// On iOS, tapping "Done" on the keyboard only fires a blur event, not a
+// keypress event, so we also attempt to submit here. The submittedViaKeyboard
+// flag prevents a double submission when Enter is used on desktop/physical
+// keyboards (which removes the input from the DOM and triggers blur).
+const handleBlur = () => {
```
cole-adams: Is there a way to do this with a form submit instead of on blur? I personally sometimes blur the text field on desktop before submitting so this change would affect current behaviour

cole-adams (approved):
Code looks fine, I guess just double check it works on iOS

PR #332, opened 2026-04-07
feat: Support arbitrary club lists

brian-norman:
## Summary
- Schema migration enabling multiple user-defined lists per club (replaces hardcoded watchlist/backlog)
- Rewrites `ListRepository` and `/club/list` handler around the new model, with Zod validation helpers
- Reworks `WatchListView` into a single page with collapsible `ListItems`, plus a `useCollapsedLists` composable for persisted UI state

## Test plan
- [ ] `npm run migrate:dev` applies `20260407_ArbitraryClubLists` cleanly on a freshly spawned dev DB
- [ ] `npm run type-check` / `npm run lint` / `npm test`
- [ ] Create, rename, reorder, and delete lists in a club; add/remove movies
- [ ] Verify collapsed state persists across reloads
- [ ] Regression-check review flow + statistics (touched composables)


migrations/schema/20260407_ArbitraryClubLists.ts
```diff
@@ -0,0 +1,191 @@
...
+
+  // 2. Add the new column (nullable for now so we can backfill).
+  await db.schema
+    .alterTable("work_list")
+    .addColumn("system_type", sql`work_list_system_type`)
+    .execute();
+
+  // 3. Backfill system_type from the old `type` enum.
+  await sql`
+    UPDATE work_list
```
cole-adams: Would prefer we use the `withTables` kysely util instead of hand-writing sql. Using kysely queries is more type safe

netlify/functions/club/list.ts
```diff
@@ -1,55 +1,115 @@
...
+  const workList = await getReviewList(clubId);
+  return res(ok(JSON.stringify(workList)));
+});
+
+// ---------------------------------------------------------------------------
+// User lists collection
+// ---------------------------------------------------------------------------
+
+router.get("/", async ({ clubId, event }, res) => {
+  const includeSystem = event.queryStringParameters?.includeSystem === "true";
```
cole-adams: Don't see a need for this? Why not just have this endpoint be "non-system lists?"

netlify/functions/utils/validation.ts
```diff
@@ -13,6 +15,11 @@ export type ClubRequest<T extends Request = Request> = T & {
   clubSlug: string;
 };
 
+export type ListRequest<T extends ClubRequest = ClubRequest> = T & {
+  listId: string;
+  listSystemType: "reviews" | "award_nominations" | null;
```
cole-adams: Define a type somewhere so we don't have to update this everywhere if we define a new list

netlify/functions/utils/validation.ts
```diff
@@ -42,3 +49,41 @@ export const validClubSlug: MiddlewareCallback<Request, ClubRequest> = async (
...
+ */
+export const validListId: MiddlewareCallback<ClubRequest, ListRequest> = async (
+  req,
+  res,
+) => {
+  const listId = req.params.listId;
+  if (!hasValue(listId)) {
+    return res(notFound("List not found"));
+  }
+  const list = await ListRepository.getListById(listId);
```
cole-adams: I would much prefer this repository method just have the club id as a parameter so we're never pulling data from a different tenant

netlify/functions/utils/validation.ts
```diff
@@ -42,3 +49,41 @@ export const validClubSlug: MiddlewareCallback<Request, ClubRequest> = async (
...
+};
+
+/**
+ * Non-generic alias of `secured` typed for ListRequest. The Router's chain
+ * inference fails to instantiate `secured`'s `T extends ClubRequest` generic
+ * to ListRequest when chained after validListId — it widens to ClubRequest
+ * and the handler loses access to listId/listSystemType. This wrapper pins
+ * T=ListRequest so the chain types resolve cleanly.
+ */
+export const securedList: MiddlewareCallback<
```
cole-adams: minor: This works, but could do the same as the awards handlers and use `secured<ListRequest>`

netlify/functions/repositories/ListRepository.ts
```diff
@@ -1,10 +1,131 @@
...
+    return db
+      .updateTable("work_list")
+      .set({ title })
+      .where("id", "=", listId)
+      .execute();
+  }
+
+  async deleteList(listId: string) {
+    // The fk_work_list_item_list_id constraint cascades work_list_item rows.
+    // The route layer must guard against deleting system lists (especially
```
cole-adams: Actually crazy to say a different part of the code should do this check. Why don't we just do it here? Its a valid concern

netlify/functions/repositories/ListRepository.ts
```diff
@@ -1,10 +1,131 @@
...
+      .where("system_type", "=", systemType)
+      .executeTakeFirstOrThrow();
+    return row.id;
+  }
+
+  async getReviewsListId(clubId: string) {
+    return this.getSystemListId(clubId, WorkListSystemType.reviews);
+  }
+
+  async getAwardNominationsListId(clubId: string) {
```
cole-adams: Honestly lets just get rid of `awards_nominations` for now. Will revisit if/when we build out awards for the masses

netlify/functions/repositories/ListRepository.ts
```diff
@@ -166,34 +276,133 @@ class ListRepository {
...
+      // Verify every id belongs to this club — prevents a malicious caller
+      // from reordering another club's lists.
+      const owned = await trx
+        .selectFrom("work_list")
+        .select("id")
+        .where("club_id", "=", clubId)
+        .where("id", "in", listIds)
+        .execute();
+
+      const ownedIds = new Set(owned.map((r) => String(r.id)));
```
cole-adams, line 329: Lets do another check here to make sure the set of ids is complete. Again don't pray for it with the documentation

netlify/functions/club/reviews.ts
```diff
@@ -16,23 +15,74 @@ const router = new Router<ClubRequest>("/api/club/:clubSlug/reviews");
...
+
+const queueSchema = z.object({
+  sourceListId: z.string().optional(),
+});
+
+// Add a work to the reviews list without creating a score yet. Used by the
+// "add review" prompt to push a movie from any user list (or a search) into
+// the reviews queue.
+router.post(
+  "/:workId/queue",
```
cole-adams: What has changed so that we need this? Why can't we use the move endpoint with the review list id?

netlify/functions/club/reviews.ts
```diff
@@ -16,23 +15,74 @@ const router = new Router<ClubRequest>("/api/club/:clubSlug/reviews");
...
+      const exists = await ListRepository.isItemInList(reviewsListId, workId);
+      if (!exists) {
+        await ListRepository.insertItemInList(reviewsListId, workId);
+      }
+    }
+    return res(ok());
+  },
+);
+
+router.delete("/:workId", secured, async ({ clubId, params }, res) => {
```
cole-adams: Same here, why don't we use the list endpoint?

netlify/functions/club/reviews.ts
```diff
@@ -16,23 +15,74 @@ const router = new Router<ClubRequest>("/api/club/:clubSlug/reviews");
...
-    workId,
-  );
-  if (!isItemInList) {
-    return res(badRequest("This movie does not exist in the list"));
+  const reviewsListId = await ListRepository.getReviewsListId(clubId);
+
+  if (hasValue(sourceListId) && sourceListId !== reviewsListId) {
+    // Move from the source user list into the reviews system list.
+    // moveItem handles "already on reviews" via ON CONFLICT DO NOTHING.
+    await ListRepository.moveItem(sourceListId, reviewsListId, workId);
```
cole-adams: Pointless addition in functionality to this endpoint. Please remove

src/features/watch-list/components/ManageListsModal.vue
```diff
@@ -0,0 +1,171 @@
...
+const deletingListId = shallowRef<string | null>(null);
+const { mutate: deleteList } = useDeleteList(props.clubSlug);
+const confirmDeleteList = () => {
+  if (deletingListId.value === null) return;
+  deleteList(deletingListId.value);
+  deletingListId.value = null;
+};
+</script>
+
+<template>
```
cole-adams, line 1: All of our components have the template at the top

src/features/watch-list/composables/useCollapsedLists.ts
```diff
@@ -0,0 +1,47 @@
...
+ */
+export function useCollapsedLists(clubSlug: string) {
+  const storageKey = `movieclub:collapsedLists:${clubSlug}`;
+
+  const initial = (() => {
+    try {
+      const raw = localStorage.getItem(storageKey);
+      if (raw === null) return new Set<string>();
+      const parsed: unknown = JSON.parse(raw);
+      if (!Array.isArray(parsed)) return new Set<string>();
```
cole-adams: Use zod to do validation here instead of hand validating

src/features/watch-list/views/WatchListView.vue
```diff
@@ -1,88 +1,163 @@
...
+
+      <div class="flex flex-col gap-6">
+        <section
+          v-for="list in userLists"
+          :key="list.id"
+          class="rounded-lg border border-slate-700 bg-slate-900/40 p-3"
+        >
+          <div class="flex items-center justify-between gap-2">
+            <div class="flex min-w-0 items-center gap-2">
+              <button
```
cole-adams: Use VBtn

src/features/watch-list/views/WatchListView.vue
```diff
@@ -1,88 +1,163 @@
...
+                class="text-slate-200 hover:text-primary"
+                :aria-label="
+                  isCollapsed(list.id) ? 'Expand list' : 'Collapse list'
+                "
+                @click="toggle(list.id)"
+              >
+                {{ isCollapsed(list.id) ? "▶" : "▼" }}
```
cole-adams: Use icons not characters

cole-adams (changes requested):

migrations/schema/20260407_ArbitraryClubLists.ts
brian-norman: Done — replaced the two raw `sql` tagged-template backfills with typed Kysely queries using `db.withTables<{ work_list: MigrationWorkListTable }>()`. The `award_nominations` backfill is also removed since that enum value is being dropped entirely.

netlify/functions/club/list.ts
brian-norman: Agreed — removed the `includeSystem` query param entirely. `GET /` now always returns only non-system lists. Added a dedicated `GET /reviews-id` endpoint so the frontend can fetch the reviews list ID without needing to include system lists in the main query.

netlify/functions/utils/validation.ts
brian-norman: Done — `listSystemType` now uses the generated `WorkListSystemType` type from `lib/types/generated/db.ts` rather than an inline string union.

netlify/functions/utils/validation.ts
brian-norman: Done — `getListById` now takes `(listId: string, clubId: string)` and filters by `club_id` in the query. The post-fetch ownership check in `validListId` is gone.

netlify/functions/utils/validation.ts
brian-norman: Done — `securedList` is removed entirely. All list-scoped routes now use the TypeScript 4.7 instantiation expression `secured<ListRequest>` directly, same pattern as the awards module (`secured<ClubAwardRequest>`).

netlify/functions/repositories/ListRepository.ts
brian-norman: Done — moved the guard into `deleteList` itself with `.where("system_type", "is", null)`. System lists are now silently ignored at the DB level regardless of which caller invokes the method.

netlify/functions/repositories/ListRepository.ts
brian-norman: Done — `award_nominations` removed from the enum (now just `['reviews']`), `getAwardNominationsListId` deleted, and the `Award Nominations` entry removed from `createListsForClub`. Awards can be revisited later if we formalize that system.

netlify/functions/repositories/ListRepository.ts
brian-norman: Done — added a completeness + duplicate check: `new Set(valid).size !== owned.length`. Using a Set rather than a plain length comparison means duplicate IDs in the payload (e.g. `["a","a"]` when the club has `["a","b"]`) are correctly rejected.

netlify/functions/club/reviews.ts
brian-norman: Removed — the queue endpoint is gone. The frontend now fetches the reviews list ID from `GET /list/reviews-id` and calls `POST /:sourceListId/items/:workId/move` with the reviews list as the destination.

netlify/functions/club/reviews.ts
brian-norman: Removed — same approach: callers now use `DELETE /:reviewsListId/items/:workId` via the generic list item endpoint.

netlify/functions/club/reviews.ts
brian-norman: Removed — `sourceListId` is gone from the schema and the handler. Scoring a movie no longer implicitly moves it; the frontend calls move separately before scoring if needed.

src/features/watch-list/components/ManageListsModal.vue
brian-norman: Fixed — `<template>` now comes before `<script setup>` in ManageListsModal.

src/features/watch-list/composables/useCollapsedLists.ts
brian-norman: Done — replaced the manual `Array.isArray` + `typeof` check with `z.array(z.string()).safeParse()`.

src/features/watch-list/views/WatchListView.vue
brian-norman: Fixed — replaced `<button>` with `<v-btn>`.

src/features/watch-list/views/WatchListView.vue
brian-norman: Fixed — replaced the ▶/▼ text characters with `<mdicon :name="isCollapsed(list.id) ? 'chevron-right' : 'chevron-down'" :size="16" />`.

src/features/watch-list/components/ListItemDetailsContent.vue
cole-adams, line 1: I'm seeing a bug with this where if I open a details drawer from one list, then click a movie from a different list,  it opens the drawer either underneath or overtop the other one. I don't think we should be layering these. It should be just one for the entire page.

.claude/rules/database.md
```diff
@@ -26,11 +26,42 @@ paths:
 - `session` - User sessions (BetterAuth managed)
 - `verification` - Email verification tokens (BetterAuth managed)
 - `movie_details` - Cached movie metadata from TMDB
-- `work_list` - Generic list for reviews, watchlist, backlog, awards
+- `work_list` - Generic list table. Each club has one or more user-defined lists (free-form titles, `system_type IS NULL`) plus optional system lists (`system_type = 'reviews' | 'award_nominations'`). A partial unique index `uq_work_list_club_system_type` enforces at most one of each system list per club. The legacy `type` enum was removed in `20260407_ArbitraryClubLists`.
```
cole-adams: Outdated comment with `award_nominations`

migrations/schema/20260407_ArbitraryClubLists.ts
```diff
@@ -0,0 +1,192 @@
...
+  const typedDb = db.withTables<{ work_list: MigrationWorkListTable }>();
+  await typedDb
+    .updateTable("work_list")
+    .set({ system_type: "reviews" })
+    .where("type", "=", "reviews")
+    .execute();
+
+  // 4. Backfill missing titles using the historical default-title map so
+  //    we can flip `title` to NOT NULL.
+  await sql`
```
cole-adams: Use kysely for EVERY query in the migration. No inlining sql where all possible

netlify/functions/club/reviews.ts
```diff
@@ -25,14 +24,12 @@ router.post("/", secured, async ({ clubId, userId, event }, res) => {
...
-    clubId,
-    WorkListType.reviews,
-    workId,
-  );
-  if (!isItemInList) {
-    return res(badRequest("This movie does not exist in the list"));
+  const reviewsListId = await ListRepository.getReviewsListId(clubId);
+  const exists = await ListRepository.isItemInList(reviewsListId, workId);
+  if (!exists) {
+    await ListRepository.insertItemInList(reviewsListId, workId);
```
cole-adams: I see no reason to change this behaviour. What is the use case for calling review on a movie that's not in the reviews list?

.claude/rules/backend-architecture.md
```diff
@@ -23,10 +23,28 @@ router.get("/:clubSlug", validClubSlug, async ({ clubSlug }, res) => {
...
 - `/api/auth/*` - BetterAuth endpoints (handled automatically)
 - `/api/club/*` - All club-related endpoints
-  - `/:clubSlug/list` - Watchlist/backlog management
-    - `PUT /:type/reorder` - Reorder list items
-    - `PUT /:type/:workId/added-date` - Update item added date
+  - `/:clubSlug/list` - Arbitrary user-defined lists per club. List IDs are
+    UUIDs; the legacy `watchlist` / `backlog` enum was removed in
+    `20260407_ArbitraryClubLists`. System lists (`reviews`,
+    `award_nominations`) are filtered out by default and only returned when
+    `?includeSystem=true`.
```
cole-adams: Not true anymore

netlify/functions/repositories/ListRepository.ts
```diff
@@ -1,10 +1,125 @@
...
+import { WorkListSystemType } from "../../../lib/types/generated/db";
 import { db } from "../utils/database";
 
 class ListRepository {
-  async getListByType(clubId: string, type: WorkListType) {
+  // -- List CRUD --------------------------------------------------------
+
+  async getListsForClub(
+    clubId: string,
+    options: { includeSystem?: boolean } = {},
```
cole-adams: get rid, you only half got rid of it in the previous update

netlify/functions/club/list.ts
```diff
@@ -58,7 +122,203 @@ async function getWorkList(clubId: string, type: WorkListType) {
...
+    );
+    if (!destination) {
+      return res(badRequest("Destination list not found"));
+    }
+    // Allow moves into the `reviews` system list (this is how the per-item
+    // review button works). Block `award_nominations` — those go through
+    // the awards UI and have their own workflow.
+    if (
+      destination.system_type !== null &&
+      destination.system_type !== WorkListSystemType.reviews
```
cole-adams: This low key does nothing since reviews is the only system list type

src/service/useList.ts
```diff
@@ -20,106 +20,388 @@ import { useAuthStore } from "@/stores/auth";
...
-export function useDeleteListItem(clubSlug: string, type: WorkListType) {
+export function useDeleteReview(clubSlug: string) {
   const auth = useAuthStore();
   const queryClient = useQueryClient();
   return useMutation({
-    mutationFn: (workId: string) =>
-      auth.request.delete(`/api/club/${clubSlug}/list/${type}/${workId}`),
+    mutationFn: async (workId: string) => {
+      const { id: reviewsListId } = (
+        await axios.get<{ id: string }>(`/api/club/${clubSlug}/list/reviews-id`)
```
cole-adams: I really don't like that we have to fetch this before every one of the mutations. Why don't we just let the callsites use the `useReviewListId` hook, so we don't have to re-fetch here?

src/service/useList.ts
```diff
@@ -20,106 +20,388 @@ import { useAuthStore } from "@/stores/auth";
...
+): UseQueryReturnType<string, AxiosError> {
+  return useQuery({
+    queryKey: ["reviewsListId", clubSlug] as const,
+    queryFn: async () =>
+      (await axios.get<{ id: string }>(`/api/club/${clubSlug}/list/reviews-id`))
+        .data.id,
+  });
+}
+
+export function useCreateList(clubSlug: string) {
```
cole-adams, line 69: You should add an optimistic update here because this list is slow to appear after hitting create. Maybe add a loading animation so the user has to wait to re-order for when we have a real id. It just doesn't feel reactive enough right now

src/service/useList.ts
```diff
@@ -20,106 +20,388 @@ import { useAuthStore } from "@/stores/auth";
...
+    mutationFn: async ({
+      workId,
+      sourceListId,
+    }: {
+      workId: string;
+      sourceListId?: string;
+    }) => {
+      if (!hasValue(sourceListId)) return;
+      const { id: reviewsListId } = (
+        await axios.get<{ id: string }>(`/api/club/${clubSlug}/list/reviews-id`)
```
cole-adams: If we get rid of the need to fetch the review-id, then we can just call the normal list hooks for the review list

src/service/useList.ts
```diff
@@ -131,18 +413,63 @@ export function useReorderList(
...
+      sourceListId: string;
+      destinationListId: string;
+      workId: string;
+    }) =>
+      auth.request.post(
+        `/api/club/${clubSlug}/list/${sourceListId}/items/${workId}/move`,
+        { destinationListId },
+      ),
+    onMutate: ({ sourceListId, workId }) => {
+      queryClient.setQueryData<DetailedWorkListItem[]>(
```
cole-adams, line 445: This optimistic update should put it in the destination list as well

src/features/watch-list/views/SharedListView.vue
cole-adams, line 1: I'm seeing a bug where the "Review" button isn't working at all. It removes from the source list but doesn't add it to the review list

src/features/watch-list/components/ListItemDetailsContent.vue
```diff
@@ -112,25 +92,46 @@
...
+      <button
+        class="flex flex-1 items-center justify-center gap-2 rounded-lg bg-red-500/20 py-3 text-red-500"
+        @click="showDeleteConfirmation = true"
+      >
+        <mdicon name="delete" />
+        <span>Delete</span>
+      </button>
+    </div>
+
+    <select
```
cole-adams, line 125: It bugs me that this select doesn't have cursor: pointer

src/features/watch-list/views/WatchListView.vue
cole-adams, line 1: Where'd the search bar go?

cole-adams (changes requested):

cole-adams (approved):

PR #344, opened 2026-05-16
Extract VSideDrawer and PosterImage shared components

brian-norman:
## Summary
- Consolidates duplicated desktop drawer markup from `MovieDetailsDrawer.vue` and `ListItemDetailsDrawer.vue` into a shared `VSideDrawer` with a real slide-in/out transition, Escape-to-close, and a coupled backdrop.
- Extracts the TMDB poster `` pattern into a shared `PosterImage` that renders a 2:3 aspect-ratio skeleton (`animate-pulse`) and fades the image in on load; cleanly handles missing posters.
- Drops the `md:pr-[35vw]` grid shrink in `GalleryView.vue` and `ListItems.vue` so the drawer overlays the grid instead of reflowing it. Removes the now-unnecessary `inline: "center"` from `scrollIntoView` calls.

Note: the manual `X` close button inside the drawer was removed; close affordances are Escape, backdrop click, and any `@close` emitted from the slot content.

## Test plan
- [ ] Open a review's details on desktop — drawer slides in, grid does not reflow, image fades in.
- [ ] Press Escape, click the backdrop, and close from inside the content — all three dismiss the drawer with a slide-out animation.
- [ ] Repeat the same checks on a watch-list / backlog item.
- [ ] Mobile bottom-sheet path still works (unchanged code path, but smoke-test both features).
- [ ] Movies without a `poster_path` render the skeleton placeholder without a broken image.
- [ ] Confirm `scrollIntoView` still keeps the selected card visible after the inline-center removal (vertical centering should be enough now that the grid isn't shrinking horizontally).

### DIFF
--- a/src/common/components/PosterImage.vue
+++ b/src/common/components/PosterImage.vue
@@ -1,0 +1,32 @@
+<template>
+  <div
+    class="overflow-hidden rounded-lg bg-lowBackground"
+    :class="{ 'animate-pulse': !loaded && hasPoster }"
+    style="aspect-ratio: 2 / 3"
+  >
+    <img
+      v-if="hasPoster"
+      :src="`https://image.tmdb.org/t/p/w500/${posterPath}`"
+      :alt="alt"
+      decoding="async"
+      loading="eager"
+      class="h-full w-full object-cover transition-opacity duration-300"
+      :class="loaded ? 'opacity-100' : 'opacity-0'"
+      @load="loaded = true"
+    />
+  </div>
+</template>
+
+<script setup lang="ts">
+import { computed, ref } from "vue";
+
+import { hasValue } from "../../../lib/checks/checks.js";
+
+const props = withDefaults(
+  defineProps<{ posterPath?: string | null; alt?: string }>(),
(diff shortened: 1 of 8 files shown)


src/common/components/PosterImage.vue
```diff
@@ -0,0 +1,32 @@
+<template>
+  <div
+    class="overflow-hidden rounded-lg bg-lowBackground"
+    :class="{ 'animate-pulse': !loaded && hasPoster }"
+    style="aspect-ratio: 2 / 3"
```
cole-adams: why not tailwind?

cole-adams (approved):

src/common/components/PosterImage.vue
brian-norman: Good call — switched to `aspect-[2/3]` in 02008cd to match the sibling components (`MoviePosterCard`, `SharedReviewView`). I'd originally inlined it because Tailwind's docs lead with `aspect-square`/`aspect-video` as the named utilities and I forgot arbitrary values were on by default in v3.

PR #352, opened 2026-05-22
Add experimental AI discussion questions feature for movie reviews

brian-norman:
## Summary

- New per-club experimental feature flag (`features.discussionQuestions`, off by default) adds a shimmering "Generate discussion questions" button to the movie review drawer that calls Gemini (`gemini-3.1-flash-lite`) and renders 3–5 thought-provoking questions to spark club conversation.
- Backend proxies the call so `GEMINI_API_KEY` stays server-side; the endpoint re-checks the feature flag and rejects with `400 "Feature not enabled"` if the UI is bypassed.
- Caching is client-side only via TanStack Query's existing `createSyncStoragePersister` (1-week `maxAge` already configured in `src/main.ts`). Reopening the drawer — or hard-refreshing — hits the cache, not the API.

## Implementation notes

- **No DB migration needed.** `club_settings.value` is already JSONB and `SettingsRepository.getSettings` spread-merges over `DEFAULT_SETTINGS.features`, so existing clubs gracefully pick up the new key as `false`.
- **AI shimmer border** uses a `conic-gradient` driven by an animated `@property --ai-angle` for a smooth rotation with no jank, plus a `prefers-reduced-motion` fallback that freezes the gradient.
- **Drawer keying fix:** added `:key="movie.id"` to `MovieDetailsContent` invocations in `MovieDetailsDrawer.vue`. The composable captures props at setup time, so without keying the cached questions would persist across review switches. Aligns with the existing `.claude/rules/code-quality.md` "Prefer Keyed Components" guidance.
- **Server-side enforcement** of the flag runs before request-body parsing — a disabled flag short-circuits cheaply.
- Trusts the client to send the movie title/year in the request body rather than re-querying `movie_details` server-side. Bounded by membership + flag; worst case is questions for the wrong movie, no leakage risk.

## Local setup gotcha

`GEMINI_API_KEY` is a Netlify-console secret and is **not** synced to local `.env`. To exercise the feature locally, devs must generate their own key at https://aistudio.google.com/apikey and add `GEMINI_API_KEY=<key>` to `.env`. This is now documented in `CLAUDE.md`.

## Known limitations

- `workId` in the URL isn't validated against an actual review record server-side. Bounded by club membership and the feature flag; acceptable for an experimental feature, worth a follow-up if it goes GA.
- Gemini model is pinned to `gemini-3.1-flash-lite` via a single constant in `netlify/functions/utils/gemini.ts`; swap as needed if Google deprecates or releases newer.

## Test plan

- [ ] Add `GEMINI_API_KEY=<your key>` to local `.env`, run `netlify dev`
- [ ] Open Club Settings → toggle "AI Discussion Questions (experimental)" on → confirm the POST to `/api/club/<slug>/settings` carries `features.discussionQuestions: true` and survives reload
- [ ] Open a movie review drawer → confirm the shimmering "Generate discussion questions" button appears below the chat and above Delete/Share
- [ ] Click the button → confirm 3–5 numbered questions appear with the expansion transition
- [ ] Close the drawer + reopen the same review → questions appear instantly with **no** network call (TanStack persistence)
- [ ] Hard-refresh the page → questions still appear instantly on reopen (localStorage rehydration)
- [ ] Open review A, generate, close → open review B → confirm B starts in idle state (validates the `:key="movie.id"` fix)
- [ ] Toggle the feature off → button disappears; manually `curl` the endpoint → returns `400 "Feature not enabled"`
- [ ] Test on a movie with an unusual title (apostrophes, non-ASCII) to confirm prompt & JSON parsing are robust
- [ ] (Optional) Enable "Reduce motion" in macOS Accessibility → confirm shimmer freezes

### DIFF
--- a/netlify/functions/club/reviews.ts
+++ b/netlify/functions/club/reviews.ts
@@ -3,9 +3,11 @@ import { z } from "zod";
 import { hasValue } from "../../../lib/checks/checks.js";
 import ListRepository from "../repositories/ListRepository";
 import ReviewRepository from "../repositories/ReviewRepository";
+import SettingsRepository from "../repositories/SettingsRepository";
 import WorkCommentRepository from "../repositories/WorkCommentRepository";
 import SharedReviewService from "../services/SharedReviewService";
 import { secured } from "../utils/auth";
+import { generateDiscussionQuestions } from "../utils/gemini";
 import { badRequest, ok, unauthorized } from "../utils/responses";
 import { Router } from "../utils/router";
 import { ClubRequest } from "../utils/validation";
@@ -150,6 +152,39 @@ router.delete(
     }
     await WorkCommentRepository.deleteById(params.commentId);
     return res(ok());
+  },
+);
+
+const discussionQuestionsSchema = z.object({
+  title: z.string().min(1).max(300),
+  releaseYear: z
+    .string()
+    .regex(/^\d{4}$/)
+    .optional(),
+});
+
+router.post(
+  "/:workId/discussion-questions",
+  secured,
+  async ({ clubId, params, event }, res) => {
+    if (!hasValue(params.workId)) {
+      return res(badRequest("No workId provided"));
+    }
+
+    const settings = await SettingsRepository.getSettings(clubId);
+    if (settings.features.discussionQuestions !== true) {
+      return res(badRequest("Feature not enabled"));
+    }
+
+    if (!hasValue(event.body)) return res(badRequest("No body provided"));
+    const body = discussionQuestionsSchema.safeParse(JSON.parse(event.body));
+    if (!body.success) return res(badRequest("Invalid body"));
+
+    const questions = await generateDiscussionQuestions(
+      body.data.title,
+      body.data.releaseYear,
+    );
+    return res(ok(JSON.stringify({ questions })));
   },
 );
 
--- a/netlify/functions/club/reviews.ts
+++ b/netlify/functions/club/reviews.ts
@@ -5,6 +5,7 @@ import ListRepository from "../repositories/ListRepository";
 import ReviewRepository from "../repositories/ReviewRepository";
 import SettingsRepository from "../repositories/SettingsRepository";
 import WorkCommentRepository from "../repositories/WorkCommentRepository";
+import WorkRepository from "../repositories/WorkRepository";
 import SharedReviewService from "../services/SharedReviewService";
 import { secured } from "../utils/auth";
 import { generateDiscussionQuestions } from "../utils/gemini";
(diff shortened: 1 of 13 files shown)


netlify/functions/utils/gemini.ts
```diff
@@ -0,0 +1,71 @@
...
+        responseMimeType: "application/json",
+        responseSchema: {
+          type: "object",
+          properties: {
+            questions: {
+              type: "array",
+              items: { type: "string" },
+              minItems: 3,
+              maxItems: 5,
+            },
```
brian-norman, line 51: Try letting it send 0 and ask it to send 0 if it doesn't know the movie

netlify/functions/club/reviews.ts
```diff
@@ -153,6 +155,39 @@ router.delete(
...
+      return res(badRequest("No workId provided"));
+    }
+
+    const settings = await SettingsRepository.getSettings(clubId);
+    if (settings.features.discussionQuestions !== true) {
+      return res(badRequest("Feature not enabled"));
+    }
+
+    if (!hasValue(event.body)) return res(badRequest("No body provided"));
+    const body = discussionQuestionsSchema.safeParse(JSON.parse(event.body));
```
cole-adams: Why do we use the title/release year from the body? Could lead to users injecting titles for the prompt. Can we just fetch the work info from the work id?

src/service/useDiscussionQuestions.ts
```diff
@@ -0,0 +1,30 @@
...
+
+export function useDiscussionQuestions(
+  clubSlug: string,
+  workId: string,
+  title: string,
+  releaseYear?: string,
+) {
+  const auth = useAuthStore();
+  return useQuery<string[]>({
+    queryKey: ["discussion-questions", clubSlug, workId, title, releaseYear],
```
cole-adams: workId should imply title and release year. We shouldn't be passing them in here

cole-adams (changes requested):

netlify/functions/club/reviews.ts
brian-norman: Good call — fixed in 6f198b8. The endpoint now resolves the movie title and release year server-side from the `workId` (via `WorkRepository.getDiscussionContext`, scoped to the club) instead of trusting the request body, so the prompt can't be poisoned by client input. This also validates the work belongs to the club.

src/service/useDiscussionQuestions.ts
brian-norman: Done in 6f198b8 — dropped `title`/`releaseYear` from the composable, its query key, and the request body. Only `clubSlug` + `workId` are passed now; the backend derives the rest from `workId`.

netlify/functions/utils/gemini.ts
brian-norman: Handled in d50f7fc: the prompt now ends with "If you do not recognize this film or cannot confirm it is a real movie, return 0 questions." The Zod schema allows an empty array and the frontend renders a dedicated "couldn't recognize this movie" state when 0 come back.

cole-adams (approved):

PR #205, opened 2025-08-08
Search filters and pills

SunnyNagam:
## Overview

Introduces a unified, pill-based **search and filter bar** and uses it across the **Reviews** and **Statistics** pages, replacing the previous one-off search inputs.

## What changed

**New `SearchFilterBar` component** (`src/common/components/SearchFilterBar.vue`)
- Free-text search input plus a row of filter "pills".
- Each pill opens a popover to set a value; applied filters render as removable chips.
- Filterable fields: Genre, Average Score, Production Company, Director, Review Date, Release Date, Runtime. Number/date fields support comparators (e.g. `>`, `<`, `=`).
- Value suggestions are derived from the current dataset with frequency counts.
- Consumers can hide irrelevant pills via `exclude-filter-keys`.
- Exposes results through `v-model:filtered-data` / `v-model:has-active-filters`.

**Enhanced filtering logic** (`src/common/searchMovies.ts`)
- `filterMovies` now supports the structured filters above (field + comparator + value) in addition to free-text matching.

**Reviews page** (`src/features/reviews/views/ReviewView.vue`)
- Replaces the inline search box with `SearchFilterBar`; tests updated accordingly.

**Statistics pages** (`ScoresView.vue`, `StatisticsView.vue`)
- Adopt `SearchFilterBar`, removing the standalone `StatisticsSearchBar` component and related composable plumbing.

## Rebase note

Rebased onto `main`. The branch's original "unify search bars" commit also reworked the old `WatchList.vue` / `ClubBacklog.vue` watch-list search. Those components were removed on `main` by #332 (arbitrary club lists), which already unified the watch-list into a single search bar feeding `ListItems`. That portion was therefore superseded and dropped during the rebase; the watch-list is unchanged from `main`. The Reviews and Statistics search unification is preserved.

## Verification

- `npm run type-check` ✅
- `npm run lint` ✅
- `npm test` ✅ (76 passing)

PR #355, opened 2026-05-27
Add book clubs alongside movie clubs

brian-norman:
## Summary

Adds **book clubs** alongside movie clubs. A club's type (movie or book) is chosen at creation and is fixed thereafter. Book metadata comes from **OpenLibrary**; movies continue to use TMDB. Everything is abstracted behind a **media-provider** seam, so adding a third media type later is one migration + one provider + one registry entry.

## What changed

**Database**
- `ClubType` enum + `club.type` column (existing clubs default to `movie`)
- `WorkType` gains `book`; new `book_details` / `book_authors` / `book_subjects` cache
- Two migrations — the `ALTER TYPE … ADD VALUE 'book'` is isolated in its own migration (CockroachDB can't combine it with other DDL)

**Backend**
- `netlify/functions/utils/providers/`: `MediaProvider` interface + `getProvider()` registry, with `MovieProvider` (TMDB) and `BookProvider` (OpenLibrary)
- Repositories/handlers now return media-agnostic work rows and enrich via `provider.getExternalData`, dispatched by `clubType` (threaded through `validClubSlug`). No `movie_details` query logic remains in the generic list/review core.
- `og-image` uses the generic stored image URL

**Frontend**
- `useMediaSearch(clubType)` dispatches search to TMDB or OpenLibrary; book "Trending" tab via OpenLibrary
- Generic `Work*` components (search prompt/card, poster card/hero, description, details content/drawer); new `BookMetadataGrid` alongside the kept movie-specific `MovieMetadataGrid` / `MovieTooltip`
- Club-type selector at creation + a type badge in the club switcher
- Awards & Statistics are hidden for book clubs (nav + a `movieClubOnly` route guard); neutral copy ("Lists", "Next Up", …)

**Fixes surfaced during testing**
- Re-validate club membership after create/join to avoid a stale-cache "Club Not Found"
- Treat freshly-added (optimistic) list items as pending so they can't be acted on before they persist (was sending the `"temp"` placeholder id)

## Verification
- `type-check`, `lint`, and all **78 tests** pass
- Migrations applied to a freshly spawned dev DB; OpenLibrary search/work/editions/trending contract validated against the live API
- Manual: create a book club → search/add via OpenLibrary → review → detail drawer shows author / first published / pages / subjects / description; Awards & Statistics hidden

## Notes
- `MovieMetadataGrid`, `MovieTooltip`, `AddMovieButton`, the statistics types, and `getDetailedMovie` intentionally keep movie names — they're tied to movie-only features (awards/statistics) or the TMDB source specifically, and are selected via the `kind` discriminant.
- This is a migration PR, so the deploy preview will spawn its own per-PR database.


lib/types/movie.ts
```diff
@@ -63,32 +63,38 @@ export interface TMDBMovieData {
...
-  runtime: number;
-  spoken_languages: string[];
-  status: string;
-  tagline: string;
-  title: string;
-  video: boolean;
-  vote_average: number;
-  vote_count: number;
+  // Everything below is sourced from nullable `movie_details` columns (and the
+  // cached list/review path can predate a detail refresh), so all are optional.
```
cole-adams: What? This comment doesn't make sense. Nothing "predates" a detail refresh. Every movie should have these columns in the `MovieDetails` table

netlify/functions/club/list.ts
```diff
@@ -110,17 +112,25 @@ router.post("/", secured, async ({ clubId, event }, res) => {
...
-    title: item.title,
-    type: item.type,
-    createdDate: item.time_added.toISOString(),
-    imageUrl: item.image_url ?? undefined,
-    externalId: item.external_id ?? undefined,
-    externalData: overviewToExternalData(item),
-  }));
+  const externalIds = items.map((item) => item.external_id).filter(hasValue);
+  const externalData =
+    await getProviderForClub(clubType).getExternalData(externalIds);
```
cole-adams: Should we be deciding by club which external provider to use? The works themselves have the work type on them. Wouldn't that make more sense to fetch by? Would it ever be a usecase that we could have multiple work types in one club?

I almost wonder if we should have a workspec like
```
{
  id: string,
  type: string,
}
```

That we just pass to an "External Data" service and get the expected output

netlify/functions/repositories/WorkRepository.ts
cole-adams, line 1: Right now we have a separate `MovieRefreshRepository` that gets run on a cron job to keep the cached data fresh. Thinking we should make that generic and have it work for books too?

src/common/components/PosterImage.vue
```diff
@@ -23,9 +23,19 @@ import { hasValue } from "../../../lib/checks/checks.js";
...
-  { posterPath: null, alt: "Movie poster" },
+  { posterPath: null, alt: "Poster" },
 );
 
 const loaded = ref(false);
 const hasPoster = computed(() => hasValue(props.posterPath));
+
+// Accept either a fully-qualified URL (OpenLibrary covers, stored image_url)
+// or a bare TMDB poster path which we expand to a TMDB image URL.
+const resolvedSrc = computed(() => {
```
cole-adams: We have an `imageUrl` property on works, why don't we use that instead of guessing wether to prefix here?

src/common/components/WorkSearchPrompt.vue
```diff
@@ -87,62 +92,59 @@ const {
...
+    searchText.value.trim().length === 0 &&
+    filteredDefaultList.value.length === 0 &&
+    searchResults.value.length === 0,
+);
 
-const selectFromSearch = (movie: MovieSearchIndex) => {
-  emit("select-from-search", movie);
-};
+const hintMessage = computed(() =>
+  clubType === ClubType.book
```
cole-adams: I don't love how this if-else would grow as we keep add more club types. Possible to like have some copy provider that is keyed on club type?

src/common/clubType.ts
cole-adams, line 1: Same with this file in general. I'd almost want each club type to define a JSON of all the different copy we need or something

src/common/workDisplay.ts
```diff
@@ -0,0 +1,35 @@
...
+
+export function asBook(
+  data: DetailedWorkData | undefined,
+): DetailedBookData | undefined {
+  return data?.kind === "book" ? data : undefined;
+}
+
+/** A short, media-appropriate subtitle (release year for movies, first
+ * published year for books) for search cards and list rows. */
+export function workSubtitle(
```
cole-adams, line 60: Again, I think something like this could live in a frontend type service for each club type

src/common/filterWorks.ts
cole-adams, line 1: This should either be renamed to just search, or possibly refactor to a club type service so each club type can implement their own search

src/features/reviews/components/WorkDetailsContent.vue
```diff
@@ -411,11 +428,19 @@ const cancelDateEdit = () => {
 
 const movieTitle = computed(() => String(props.movie.renderValue("title")));
 
-const releaseYear = computed(() => {
-  const releaseDate = props.movie.original.externalData?.release_date;
-  if (!hasValue(releaseDate)) return undefined;
-  const year = DateTime.fromISO(releaseDate).year;
-  return Number.isNaN(year) ? undefined : year;
+const movieData = computed(() => asMovie(props.movie.original.externalData));
```
cole-adams, line 480: Want to avoid having to add these data properties as more works exist

src/features/reviews/views/SharedReviewView.vue
```diff
@@ -9,9 +9,9 @@
       <div v-if="data.work" class="relative overflow-hidden rounded-lg">
         <!-- Backdrop Image -->
         <div
-          v-if="data.work.externalData?.backdrop_path"
+          v-if="movieData?.backdrop_path"
```
cole-adams: Don't we want this to work for books too?

src/features/statistics/types.ts
```diff
@@ -3,7 +3,10 @@ import { DetailedMovieData } from "../../../lib/types/movie";
 
 export interface MovieData {
   id: string;
-  type: WorkType.movie;
+  // Statistics only run for movie clubs (gated in the router/nav), so this is
+  // always WorkType.movie in practice; typed as WorkType to avoid enum-literal
+  // widening friction at the mapping site.
+  type: WorkType;
```
cole-adams: Not a fan of widening a type when it will never be used that way. Why can't we just do a filter before passing to anything with movie data. A simple `isMovieData` function could correctly gate the type

src/features/watch-list/components/AddWorkModal.vue
```diff
@@ -0,0 +1,142 @@
...
+      subtitle: movie.release_date ? movie.release_date.slice(0, 4) : undefined,
+      imageUrl: movie.poster_path
+        ? `${BASE_IMAGE_URL}${movie.poster_path}`
+        : undefined,
+    })),
+  );
+});
+
+// -- Book clubs: OpenLibrary trending --
+const bookTabs: { key: BookTrendingPeriod; label: string }[] = [
```
cole-adams, line 63: Again here, would LOVE to consolidate all club type logic into central service

src/router/index.ts
```diff
@@ -46,10 +47,18 @@ const checkClubAccess = async (
...
     return next();
-  } else {
-    // User is not a member of this club
-    return next({ name: "ClubNotFound" });
   }
+
+  // The cached club list may be stale right after creating or joining a club
+  // (the invalidated query hasn't refetched yet). Re-fetch once and re-check
+  // before declaring the club inaccessible.
+  await auth.refreshClubs();
```
cole-adams, line 46: I don't like this. Can we not just include `isRefetching` in the `watiForClubsReady` call? To guarentee we always have a fresh value/

cole-adams (changes requested):

lib/types/movie.ts
brian-norman: Good catch — that parenthetical was wrong. A row only comes back from `getExternalData` when `overview` is present, and details are populated on add, so nothing "predates" anything. I rewrote the comment: the fields are optional because (1) they map from nullable `movie_details` columns (coerced `null → undefined`), and (2) a handful (`id`, `title`, `vote_count`, `video`, `spoken_languages`, `original_title`) aren't read by the list/review mapper at all — only the richer full-TMDB fetch used by awards sets them.

netlify/functions/club/list.ts
brian-norman: Done, and agreed the work's type is the better key. The registry was already keyed on `WorkType` (`getProvider(type)`); I added `getExternalDataForWorks(works: {externalId, type}[])` that groups works by their own `type`, calls each provider, and merges the maps — basically the `{id, type}` → External Data service you described. `list.ts` (list + reviews paths) and `SharedReviewService` now use it, so `clubType` is no longer threaded into enrichment. Clubs are 1:1 movie/book today, but keying on the work means a mixed-type list would just work.

netlify/functions/repositories/WorkRepository.ts
brian-norman: Done — made it generic. On-add caching already dispatched via `getProvider(work.type).fetchAndCacheDetails`; I added `refreshStaleDetails(limit)` to the `MediaProvider` interface (movie = TMDB re-fetch + `updateMovieDetails`; book = OpenLibrary re-fetch + upsert of `book_details`/authors/subjects), each ordered by oldest `updated_date`. `MovieRefreshRepository` is gone, and `scheduled-movie-refresh` → `scheduled-work-refresh`, which just sweeps `allProviders()` and aggregates results. Adding a third type needs nothing here.

src/common/components/PosterImage.vue
brian-norman: Done. `PosterImage` now takes a fully-qualified `imageUrl` and just renders it — no http-prefix guessing. Resolution moved into one helper, `workPosterUrl(externalData, fallbackImageUrl)`, which prefers the movie's TMDB poster (expanded to w500, higher-res than the w154 URL stored on the work) and otherwise uses the work's `imageUrl` (book cover, or a movie before its details are cached). `WorkDetailsContent`, `ListItemDetailsContent`, and `SharedReviewView` all use it.

src/common/components/WorkSearchPrompt.vue
brian-norman: Done — there's now a `CLUB_TYPE_CONFIG: Record<ClubType, ClubTypeConfig>` in `src/common/clubType.ts` holding icon, label, noun, `searchHint`, and the `search` impl per type. The copy if-else here is now just `clubTypeConfig(clubType).searchHint`.

src/common/clubType.ts
brian-norman: Did this — `clubType.ts` is that registry now: one `ClubTypeConfig` entry per club type with all the per-type copy/icons/search. Adding a third type is one entry here (plus the migration + provider).

src/common/workDisplay.ts
brian-norman: Kept these here but added `isMovieData`/`isBookData` guards and a `workPosterUrl` helper, with `asMovie`/`asBook` now defined in terms of the guards. These narrow the `DetailedWorkData` *union* (keyed on `kind`) — the data-level counterpart to club type — while the club-type registry proper lives in `clubType.ts`. Happy to fold them together if you'd prefer one module.

src/common/filterWorks.ts
brian-norman: Two parts: search now lives in the club-type registry (`clubTypeConfig(type).search`, with `searchMovies`/`searchBooks` in `clubType.ts`) and `useMediaSearch` just dispatches through it. This file was actually the *filter* function, so I renamed it `searchMovies.ts` → `filterWorks.ts` / `filterWorks`.

Heads-up from rebasing onto `main`: the new `SearchFilterBar` (#205) now drives review/list/stats filtering, replacing the old `searchTerm`-string path. I wired `filterWorks` into it and made it union-safe (`asMovie`/`asBook`), and `filterWorks` still supports the book `author`/`subject` filters — but `SearchFilterBar`'s filter *options* are still movie-only (genre/director/company/…), so book clubs don't yet get author/subject pills (and will see movie pills with empty suggestions). It already has an `excludeFilterKeys` prop, so making it club-type-aware is a tractable follow-up; flagging it rather than expanding this PR further.

src/features/reviews/components/WorkDetailsContent.vue
brian-norman: Moved the year fork into the shared `workSubtitle(externalData)` helper, so `displayYear` here (and in `ListItemDetailsContent`) just calls it instead of inlining movie/book field access. Type-specific metadata still renders via the `MovieMetadataGrid`/`BookMetadataGrid` split gated by `asMovie`/`asBook`.

src/features/reviews/views/SharedReviewView.vue
brian-norman: Books don't have a backdrop — OpenLibrary only provides a cover image, no wide backdrop art — so the backdrop stays movie-only (gated via `asMovie`/`movieData`) and the hero just renders without it for books. The poster/cover itself works for both now via `workPosterUrl`. (Same reasoning for the TMDB external links + WatchProviders in the detail drawer — I gated those with `v-if="movieData"` so they don't show for book clubs.)

src/features/statistics/types.ts
brian-norman: Reverted `type` back to the `WorkType.movie` literal and added `isMovieData` (in `workDisplay`); the mapper gates with `if (!isMovieData(externalData)) return null`. The "widening friction" was just the `.map` callback's inferred return widening `WorkType.movie` → `WorkType` — annotating the callback return as `MovieData | null` gives it contextual typing so the literal sticks, no interface widening needed.

src/features/watch-list/components/AddWorkModal.vue
brian-norman: Pulled the per-type copy/icon/search/`workType` into the `clubType.ts` registry (used here via `workTypeForClub` + the shared `WorkSearchPrompt`). The two "default list" pipelines stay split because they're structurally different composables — TMDB collections are paginated (`useInfiniteCollection` + load-more) and OpenLibrary trending is a flat list — so they can't collapse into one config field without fighting the hook rules. Open to ideas if you'd like them unified further.

src/router/index.ts
brian-norman: `waitForClubsReady` now waits on `isFetching` (not just `isLoading`), so the refetch that `invalidateQueries(['user','clubs'])` kicks off on create/join is awaited there — that's the core of your suggestion. I kept a single forced `refreshClubs()` on the membership-*miss* path as a safety net: `useCreateClub` fires the invalidation un-awaited in `onSuccess`, so if that refetch hasn't flipped `isFetching` to true by the time the guard checks, the `isFetching` wait alone would race and we'd flash "Club Not Found" again. The miss-path refetch only runs when the first check fails, so it's not the every-navigation round-trip the old code did. Happy to drop it if you'd rather rely on `isFetching` alone.

brian-norman:
Thanks for the thorough review @cole-adams — addressed all of it (replied inline on each thread). Highlights:

- **Central club-type registry** (`src/common/clubType.ts`): `CLUB_TYPE_CONFIG` keyed by `ClubType` owns per-type copy, icons, noun, and the `search` impl. `WorkSearchPrompt`, `AddWorkModal`, `ClubSwitcher`, and `useMediaSearch` all flow through it.
- **Enrich by the work's `WorkType`, not the club** — `getExternalDataForWorks` groups works by `type` and merges per-provider maps (`list.ts`, reviews, shared review).
- **Generic refresh** — `MediaProvider.refreshStaleDetails` + `scheduled-work-refresh` sweeping `allProviders()` (replaces the movie-only `MovieRefreshRepository`).
- `PosterImage` takes a resolved `imageUrl` (`workPosterUrl`); `isMovieData`/`isBookData` guards; statistics un-widened; router folds `isFetching` into `waitForClubsReady` (kept a miss-path refetch as a safety net).

Also **rebased onto `main`** and resolved conflicts. The branch is squashed to a single commit (it was one commit anyway). Notable integration: `main`'s new `SearchFilterBar` (#205) now drives filtering — I wired the renamed, union-safe `filterWorks` into it. One follow-up I deliberately left out of this PR: surfacing book `author`/`subject` filter pills in `SearchFilterBar` for book clubs (the function supports them; the UI options are still movie-only — `excludeFilterKeys` makes it club-aware-able). Called out on the search thread.

`type-check` / `lint` / **83 tests** all green.

cole-adams (approved):

PR #358, opened 2026-06-02
Add shareable club statistics page

brian-norman:
## What

Makes a club's Statistics page shareable via a public link that anyone can open without an account — same pattern as the existing shared review/list pages — simplifies the Statistics page down to a single insights view, and **unifies the chrome across all three public `/share` pages** with one shared header component.

- **New public route** `/share/club/:clubSlug/statistics` (`noAuth: true`), so logged-out visitors aren't bounced to ClubNotFound.
- **New `SharedStatisticsView.vue`** that renders the insights stats wrapped in shared-page chrome and the "Join the Club!" CTA banner for logged-out visitors.
- **Share button** added to the in-app Statistics page header, wired to the existing `useShare` composable (native mobile share + clipboard fallback + toast).
- **Removed the table (scores) view** from both the in-app and shared Statistics pages, leaving the insights dashboard as the only view. Drops the dashboard/table toggle and the now-unused `ScoresView`, `useScoresTable`, and `StatisticsSearchBar`, plus the dead search/score-context state in `useStatisticsData`.

## Shared-page chrome

The branded hero is extracted into one reusable **`SharedPageHeader.vue`** (MovieClub wordmark, gradient background, club-name title, optional eyebrow/subtitle, and an overlapping member-avatar row with a `+N` overflow chip) and used by all three public pages so they look consistent by construction.

- **Removed the large single club avatar** from the stats hero; the member-avatar row is kept.
- **Shared review page** now shows the branded header above the existing movie backdrop/poster hero.
- **Shared watch-list page** now shows the branded header (with the list title as a subtitle) and member avatars; it fetches members via the public `useMembers`.
- The "Join the Club!" CTA banner was already present on all three shared pages — unchanged.

`SharedPageHeader`'s `members` prop is typed to the minimal `{ id, name, image? }` shape it actually reads, so it accepts both the full `Member[]` (stats/list) and the trimmed member shape from the review endpoint with no casts.

## Why no backend changes

The Statistics page computes everything on the frontend from three endpoints that are already public (gated only by `validClubSlug`, no auth): `GET /api/club/:slug`, `/members`, and `/list/reviews`. Sharing stats exposes nothing the API doesn't already serve unauthenticated. The watch-list page's new member-avatar row uses the same already-public `/members` endpoint.

## Testing

- `npm run type-check`, `npm run lint`, `npm test` (80 passed), and `vite build` all pass.
- Manual: open the share URL in a logged-out window → insights stats render with the branded hero and CTA banner, no redirect. Empty-review club shows the "No Statistics Yet" state.
- Note: the consistent rendering of the three shared pages should be eyeballed in a browser (`netlify dev` → the three `/share/club/...` URLs logged out), including the review page's stacked branded-header-above-movie-hero layout.


src/common/components/SharedPageHeader.vue
```diff
@@ -0,0 +1,76 @@
+<template>
+  <header
+    class="bg-gradient-to-b from-secondary/40 to-transparent px-6 pb-10 pt-6"
```
cole-adams: ngl I low key hate the gradient because there's a hard normal background above

cole-adams (changes requested):

src/common/components/SharedPageHeader.vue
brian-norman: Dropped the gradient entirely (ea96d81) — the header now sits flat on the page background, consistent with the navbar above it. No more hard edge.

cole-adams (approved):

PR #368, opened 2026-06-06
Add member attribution to list items

brian-norman:
Track which member adds a movie to a list via a new nullable
added_by_user_id column on work_list_item, and surface it on the
lists page as an avatar + name on each card and an "Added by X on
[date]" line in the item details drawer.

- Migration adds added_by_user_id (FK to user, ON DELETE SET NULL)
- insertItemInList records the current user; moveItem preserves the
  original adder when an item is moved between lists
- getListItems returns the adder as addedBy on WorkListItem; the
  frontend resolves it to a club Member via useMembers
- Existing items have no recorded adder and render without attribution

### DIFF
--- a/lib/types/generated/db.ts
+++ b/lib/types/generated/db.ts
@@ -243,6 +243,7 @@ export interface WorkList {
 }
 
 export interface WorkListItem {
+  added_by_user_id: Int8 | null;
   list_id: Int8;
   position: Generated<Int8>;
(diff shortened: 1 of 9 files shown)


cole-adams (commented):
Looks good, just need a rebase @brian-norman

cole-adams (approved):

PR #374, opened 2026-06-10
Make search filters club-type-aware

brian-norman:
The shared SearchFilterBar offered a hardcoded set of movie-only filters
(Genre, Production Company, Director, Release Date, Runtime), which made no
sense for book clubs.

Filter options now flow from the CLUB_TYPE_CONFIG registry: each club type
declares its own filterOptions, and SearchFilterBar takes a clubType prop and
builds both the pills and the enum suggestions generically from that registry.
Book clubs get Author, Subject, First Published, and Pages (plus the shared
Average Score / Review Date); movie clubs are unchanged.

- Add FilterOption + per-type filterOptions to clubType.ts, with an enum
  `suggestions` selector so suggestion aggregation is no longer movie-specific.
- Add first_publish_year and pages numeric filters to filterWorks; remove the
  now-resolved TODO and refresh the doc comment.
- SearchFilterBar consumes the registry and drops the hardcoded movie options
  and asMovie-based suggestion counts.
- ReviewView/WatchListView pass the club type; ReviewView's empty-state copy is
  now club-type-aware.
- Add filterWorks tests for the new book numeric comparators.

https://claude.ai/code/session_01XEJDSZ5rf6KmNm1P7eryMx

### DIFF
--- a/src/common/filterWorks.ts
+++ b/src/common/filterWorks.ts
@@ -1,30 +1,26 @@
-import { DetailedBookData } from "../../lib/types/book";
-import { DetailedWorkListItem } from "../../lib/types/lists";
-import { DetailedMovieData } from "../../lib/types/movie";
-
-import { asBook, asMovie } from "@/common/workDisplay";
+import { hasValue } from "@/../lib/checks/checks";
+import { ClubType } from "@/../lib/types/generated/db";
+import { DetailedWorkListItem } from "@/../lib/types/lists";
+import { clubTypeConfig } from "@/common/clubType";
+import { FilterQuery, includesCaseInsensitive } from "@/common/filterMatchers";
 
 /**
- * Filters work/review rows by structured filters and optional title text.
+ * Filters work/review rows down to those matching every applied structured
+ * filter plus the optional free-text title search.
+ *
+ * Which filters exist — and how each one matches — is defined entirely by the
+ * club type's `filterOptions` registry (see `clubType.ts`). This function is
+ * media-agnostic: it walks that registry and delegates to each applied
+ * option's `matches` predicate, so it never grows a branch when a new club
+ * type or field is added.
  *
  * @param works - The rows to filter.
- * @param searchQuery - An object with:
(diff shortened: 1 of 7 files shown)


src/common/filterWorks.ts
cole-adams, line 1: I don't like how this method will keep growing as we add more club types. Can we refactor so the filtering logic lives in the club type config as well?

src/features/reviews/views/ReviewView.vue
```diff
@@ -164,6 +165,17 @@ const showEmptyState = computed(
   () => !loading.value && filteredReviews.value.length === 0,
 );
 
+const isBookClub = computed(() => club.value?.type === ClubType.book);
+const searchEmptyDescription = computed(() =>
+  isBookClub.value
```
cole-adams: da fuq, you're literally using the club type config below. This type of conditional is EXACTLY what the club type config is supposed to prevent

cole-adams (changes requested):

src/common/filterWorks.ts
brian-norman: Done in 768b82e — `filterWorks` no longer has a branch per field/club type. Each `FilterOption` in `CLUB_TYPE_CONFIG` now owns a `matches` predicate, and `filterWorks` just walks the active club type's `filterOptions` and delegates (it takes a `clubType` arg to select the registry). The comparator/matcher primitives moved to a shared `filterMatchers.ts`. A nice side effect: filtering is now scoped to the type — a book club can no longer be filtered by a movie-only key like `genre`. Adding a club type or field is a registry-only change from here.

src/features/reviews/views/ReviewView.vue
brian-norman: Fixed in 768b82e — dropped the `isBookClub` ternary. Added a `searchableFieldsHint` field to the registry, so this now reads `clubTypeConfig(club.value?.type ?? ClubType.movie).searchableFieldsHint` like `noReviewsDescription` right below it.

cole-adams (approved):

PR #375, opened 2026-06-11
Test suite overhaul: backend testability, strict MSW, UI coverage, coverage gates

brian-norman:
## Summary

Comprehensive test-suite upgrade: the backend is now testable at all, MSW is strict, and the suite grew from **83 tests in 6 files to 1315 tests in 123 files** — all green, with `type-check`, `lint`, and `format:check` clean and coverage thresholds guarding regression.

> **Merged with `main` through #447.** See [Reconciliation with main](#reconciliation-with-main).
>
> **Backend tests are integration tests** against a real CockroachDB, following @cole-adams's review. See [Backend: integration tests](#backend-integration-tests).

### Testing infrastructure
- **Vitest split into three projects** (`vite.config.ts`): `client` (jsdom + MSW), `server` (node, pure units), and `integration` (node, real database). Previously `test.root: "src/"` made `netlify/functions/` and `lib/` invisible to the runner — zero backend tests could exist.
- **MSW strict mode** in both the client and integration projects: `onUnhandledRequest: "error"` — unhandled requests now fail tests instead of silently leaking to the network. Added the missing `reviews-id` and `reviews/:workId/comments` baseline handlers; per-test handlers use `server.use(...)`.
- **Coverage** now spans `src/`, `lib/`, and `netlify/functions/` (**82.2% statements**, vs. an effective ~24% before when the backend was uncounted) with istanbul **thresholds (82/70/77/83)** that fail the build on regression. Coverage output moved out of `src/coverage` (gitignore updated).
- **oxlint test-file override**: `oxlint.config.ts` disables `typescript/unbound-method` for `*.test.ts`/`*.spec.ts` only. (This replaces the ESLint override this PR originally carried; #434 deleted `eslint.config.mjs`, and oxlint does not enable the other two rules at all.)
- **No `as` casts**: the suite honours the repo's no-`as` convention end to end — wide unions (ag-charts options) go through runtime predicate helpers, JSON boundaries through the test HTTP client, and runtime-only properties through `toMatchObject`.
- `.claude/rules/testing.md` rewritten on top of main's version: the three projects, strict-MSW workflow, the integration harness, and the client gotchas that each cost an hour.

### Backend: integration tests

`netlify/functions/tests/` runs every Netlify function handler end to end against a live CockroachDB started with `@testcontainers/cockroachdb`.

- **One container per run.** `setup/globalSetup.ts` boots it and migrates it from scratch through the same `tsx ./migrations/schemaMigrator.ts` the deploy runs — a second copy of that wiring in the test tree would drift. Startup + migrations cost ~45s, paid once.
- **Isolation is a `DELETE` across the domain tables between tests** (~15ms). Cockroach implements `TRUNCATE` as a schema change costing ~750ms per call, which is a non-starter at this test count. The auth tables survive on purpose: signing a user up costs two bcrypt hashes, so three fixture users are created once per file and reused, with profile edits undone between tests.
- **What is real:** the routers, `validClubSlug` / `validListId`, BetterAuth's `loggedIn` / `secured`, every repository, Kysely, the Cloudinary SDK, and the migrated schema. `signIn("alice")` runs `auth.api.signUpEmail` + `signInEmail` and captures the actual signed cookie, so authorization is checked against real `session` rows.
- **What is faked:** the third-party HTTP APIs, and only at the network boundary (MSW) — TMDB, Google Books, Gemini, Resend, Cloudinary. `utils/tmdb.ts`, `providers/googleBooks.ts`, `utils/gemini.ts` and `utils/email.ts` all execute their real request-building and response-parsing.
- **Coverage:** `netlify/functions` went from 63% to **95%** statements — repositories 90%, routes 93%, utils 96% — without a single unit test for a repository, because the handlers exercise them for real.

Routes covered: club CRUD and slug management, lists (positions, system-list protections, the move transaction, the add-work upsert and its provider cache), reviews and comments with every authorization path, per-work scores, discussion questions, the shared-review payload, members and the invite/join flow, settings, the awards sub-router, `member.ts` (including multipart avatar upload through the real Cloudinary SDK), `og-image`, `auth.ts`, and both scheduled functions — `scheduled-db-cleanup` creates and drops real throwaway databases on the container.

Helpers live in `netlify/functions/tests/`: `helpers/http.ts` (`requester(handler)` → `.get/.post/.put/.delete`), `helpers/auth.ts` (real sessions), `helpers/factories.ts` (row seeding, deliberately not routed through the repositories so a test's arrange step never depends on the code it is about to exercise), `helpers/database.ts`, `setup/externalApis.ts` and `fixtures/external.ts`.

**`npm test` now requires Docker.** `npm run test:unit` runs the client and pure-backend projects without it; `npm run test:integration` runs only the database-backed ones. CI is unaffected (`ubuntu-latest` ships a daemon) and so is Netlify, whose build command never ran the tests.

### Backend: unit tests (`server` project)

Pure functions with no database — `lib/checks`, `lib/googleBooks`, and `netlify/functions/utils/` (router, responses, slug, tmdb, gemini, email, movieDetailsUpdater, providers' mappers). These may mock the transport (axios, Resend), because there is nothing else for them to reach.

### Frontend tests (new)
- All service composables (`useClub`, `useList`, `useReviews`, `useAwards`, `useUser`, `useTMDB`, `useDiscussionQuestions`), auth store, common composables, LazyLoad
- Common components: VTable, CommentThread, VSideDrawer, MovieDescription, MovieMetadataGrid, PosterImage, and more
- Feature views/components: auth views, clubs views, **profile** (ProfileView), **settings** (ClubSettingsView, ManageListsModal), **watch-list** (WatchListView, ListItems, AddWorkModal, ListItemDetails drawer/content), **awards frontend** (AwardsView, NominationsView, RankingsView, ResultView, CategoriesView, YearView), shared review/list views, and statistics

### Statistics and reviews UI coverage

The statistics feature is covered end to end — all 17 widget components, both shared controls, and all three views (`StatisticsView`, `InsightsView`, `SharedStatisticsView`) — along with the reviews surfaces this PR previously left open (`GalleryView`, `DiscussionQuestions`, `AddReviewPrompt`). The `src/features/statistics` tree sits at **95.5% statements**.

Two pieces of infrastructure unblocked it:

- **`src/mocks/agCharts.ts`** stubs `ag-charts-vue3`. The real component paints into a `<canvas>` jsdom has no 2D context for, and throws on both mount (`ModulesManager.addModule`) and unmount (`Scene.strip` → `ctx.setTransform`). Pulling in the native `canvas` addon to paint pixels no test can read would be pure cost, so specs opt in with `vi.mock("ag-charts-vue3", async () => await import("@/mocks/agCharts"))` — the dynamic import sidesteps the TDZ error a statically-imported stub hits under `vi.mock` hoisting. The stub renders `role="img"` and exposes `chartOptions(el)` / `chartSeriesNames(el)`, so widget specs assert that switching mode *reconfigured* the chart while the option builders stay covered in detail by `chartOptions.spec.ts`.
- **`src/features/reviews/tests/reviewTable.ts`** provides `withReviewTable()`, which wraps a component in a host that builds a real TanStack table over fixture reviews. `useVueTable` needs a component instance, so the table the gallery takes as a prop cannot be constructed in a spec's module scope.

Supporting changes:

- `mockIntersectionObserver` gains an opt-in `{ intersecting: true }` mode. It previously always recorded `observe` without firing, which leaves a `v-lazy-load` image's real URL parked in `data-src` with an empty `src` — fine for most specs, but it makes the resolved `src` unassertable. Default behaviour is unchanged for the 20 existing callers.
- The statistics `WorkStatsData` / `Member` factories move out of `statsComputers.test.ts` into a shared `tests/fixtures.ts` (its tests pass unchanged), alongside a `SCORED_MOVIE_REVIEW` payload the view specs serve.
- `.claude/rules/testing.md` documents the AG Charts stub, the observer modes, the review-table harness, and two traps worth writing down: `userEvent.setup()` installs its own `navigator.clipboard`, so a clipboard spy stubbed in `beforeEach` is silently replaced (the test fails with "0 calls" while a success toast proves the copy happened); and `computeScoreTrend` reads `work.scores`, not `work.userScores`, while both rolling charts clamp their window to `Math.max(5, …)` — so trend/spread need six scored reviews before they plot anything.

## Reconciliation with main

**Earlier rebase.** `main` moved 47 commits while this branch was open (through #435): `gemini.ts` became a generic `generateJson` with prompt wording per `MediaProvider`; list adds became an upsert (#421); `blurScores` (#383) and `legacy_id` (#428) are gone; movie payloads split into summary (`castNames`) vs detail (`actors`) (#421); awards nominations/rankings are keyed by **user id** rather than name (#397) — with a test that a rename follows through to the nominator avatars, which is the bug #397 fixed; and `useAllUserListItems` fetches one aggregated `/list/all-items` endpoint instead of fanning out per list.

Tests replaced along the way because the feature was redesigned: `ScoreEntryPanel` superseded the old score-entry flow (#407); `MovieMetadataGrid` relabelled its fields; the `normalizeArray` / `getScoreContextColor` scoring tests went with the functions #412 deleted; and the `useUpdateClubSettings` "optimistic update" test asserted nothing (its `waitFor` was `.catch()`-swallowed and the ref it checked was never wired up) — replaced with real optimistic-update and rollback-on-error tests.

**Merge through #447.**
- Main removed the reviews table view (#431), so the `TableView`, `ReviewScore` and `MovieTooltip` specs go with their components.
- Main added `majorCastNames` and `TMDBCastMember.popularity`; every movie fixture, backend and frontend, carries them now.
- `getErrorMessage` was deleted (#447); its tests go with it. `validClubSlug` / `validListId` are still covered, now against real rows.
- Main stubs `HTMLElement.prototype.scrollIntoView` globally in `src/tests/setup.ts`, which shadowed `GalleryView.spec`'s own `Element.prototype` stub — the spec asserts against the global one now.
- Main's watchlist toast (#436) put a second button in the DOM for the `useList` mutation harnesses; they query their button by name now.
- Main rewrote `CLAUDE.md` and `.claude/rules/` (#437); this branch's testing docs were rewritten on top of that style rather than reverting it.

## New coverage for main's code

- **`lib/googleBooks.ts`** (#396) — cover selection and the varchar(255) drop, published-year parsing, HTML/entity stripping, popularity sort stability, BISAC category flattening.
- **`providers/googleBooks.ts`** — volume fetch and search; key present vs absent (an empty `key` param is a 400), search defaults, absent `items` on no results.
- **`utils/reviewScores.ts`** (#421) — departed members excluded from both the map and the average; null outer-join rows ignored.
- **`statistics/statisticsWidgets.ts`** (#412) — the club-type widget registry: every type covered, keys unique, media-specific widgets stay on their own type, member-count visibility gates.
- **`statistics/useStatisticsData.ts`** — books count without metadata, movies don't; unscored works dropped; the per-member histogram.
- **Chart options + palette** — stacked per-member series, bins recounted from filtered works, the 0–10 decade axis, compact mode.
- **`MemberFilterChips` / `SegmentedToggle`**, **`useShare`** (including the iOS omit-text fix from #415), **`useRandomPicker`**, and **`watchUntil`**.

### Determinism
- **Timezone pinned to UTC** (`test.env = { TZ: "UTC" }`). Components format date-only strings with `new Date(...)` — parsed as UTC midnight — then render with `toLocaleDateString`, so west of Greenwich a release date renders as the previous day. CI (UTC) and a Pacific dev machine disagreed on exactly one assertion; the pin makes every date assertion mean the same thing wherever the suite runs (`TZ=Asia/Tokyo npm test` stays green).
- **CockroachDB image pinned** to `v25.3.3` (override with `COCKROACH_TEST_IMAGE`). v25 is the floor: on v24.1 the `20260407_ArbitraryClubLists` migration fails from scratch, because Cockroach still reports `work_list` as depending on `work_list_type` while the `DROP COLUMN` schema-change job is in flight.

### Production changes (deliberate, minimal)
- **`club/awards/{category,nomination,ranking,step}.ts`: bug fix.** These four sub-routers still declared `/api/club/:clubId<\d+>/awards/…` base URLs from before the slug migration, so the sub-router's own path test never matched a real (non-numeric) club slug and **every awards write returned 404 in production**. The frontend calls `/api/club/${clubSlug}/awards/…`. The mocked tests hid this by requesting `/api/club/1/…` to satisfy the numeric constraint; the integration tests failed on it immediately. Base URLs now use `:clubSlug`.
- `ListItems.vue`: **bug fix** — the watch-list empty state passed `header`/`message` to `<empty-state>`, which reads `title`/`description`, so it rendered blank. Surfaced by the new test.
- `watchUntil.ts`: **bug fix** — with `immediate: true`, a predicate that already held ran before `const unwatch` was assigned and threw a TDZ `ReferenceError`. Every current caller guards the immediate case with an `if`, so it was latent, but the helper's contract says otherwise. Now uses a nullable handle and unregisters after assignment.
- `CommentThread.vue`: `aria-label`s on three icon-only buttons (edit/delete/send) — accessibility fix that also makes them queryable by role.
- `movieProvider.ts`: `toMovieDataSummary` exported (with a doc comment) so the row→`MovieDataSummary` mapping can be unit-tested directly.
- `src/tests/setup.ts`: global router mock's `push()` returns a Promise, matching the real Vue Router contract; MSW listens with `onUnhandledRequest: "error"`.

### Known follow-ups
- **`og-image.ts` reports "from 1 rating" for an unscored work.** The rating count comes from the raw query rows, and `getReviewsByWorkId` left-joins `review`, so a work nobody has scored still yields one all-null row. The integration test pins the current behaviour with a comment rather than fixing it, to keep the production diff in this PR to the awards 404. Counting `scores` instead of `reviews` is the fix.
- **`WidgetShell`'s `innerClass` / `outerClass` props have no callers** anywhere in `src/`. Rather than write CSS-selector tests for dead API, they are left untested; deleting them is a one-line cleanup for a non-test PR.
- **`src/mocks/data/reviews.json` predates the `kind` discriminant** on `externalData`, so `isMovieData` rejects every review in it and `useStatisticsData` maps the whole fixture to zero works. The statistics view specs serve their own payload instead of editing the shared fixture; backfilling `kind` there would make the baseline handler faithful to real responses.
- **`GalleryView` passes `null` to `WorkPosterCard`'s `String`-typed `posterUrl`** for a review with no image — TanStack's `renderValue` falls back to `null`, not `undefined` — which logs a Vue prop-type warning. Cosmetic, and the placeholder still renders.


cole-adams:
I understand the usefulness of this PR, but it looks like the build time when from ~2m40s to 4m. Given our current build minute concerns, should we move forward with this? Would reduce the number of builds we can perform per month.

Maybe we wait until the end of the month to merge this to see if the github merge change frees up enough time?

cole-adams (changes requested):
I'll be honest, for the backend tests, I would much prefer a more integration-style format for these tests, mocking as few things as humanly possible.

I would achieve this by using the `@testcontainers/cockroachdb` which would allow us to essentially spin up an in-memory version of our DB, then we can test the backend end-to-end without having to mock a bunch of data

brian-norman:
Done — the backend tests are now integration tests against a real CockroachDB via `@testcontainers/cockroachdb`. Pushed in bd97aee.

**Setup.** A third Vitest project, `integration`, boots one throwaway CockroachDB in `globalSetup` and migrates it from scratch through the same `tsx ./migrations/schemaMigrator.ts` the deploy runs. Isolation between tests is a `DELETE` across the domain tables (~15ms — Cockroach's `TRUNCATE` is a 750ms schema change, so it was a non-starter at ~250 tests).

**Real:** the routers, `validClubSlug`/`validListId`, BetterAuth's `loggedIn`/`secured`, every repository, Kysely, the Cloudinary SDK, and the migrated schema. Tests sign in through `auth.api.signUpEmail`/`signInEmail` and send the actual signed cookie, so authorization is verified against real `session` rows rather than a pass-through stub.

**Faked:** the third-party HTTP APIs only, and at the network boundary (MSW) — TMDB, Google Books, Gemini, Resend, Cloudinary — with `onUnhandledRequest: "error"`. That keeps `utils/tmdb.ts`, `providers/googleBooks.ts`, `utils/gemini.ts` and `utils/email.ts` executing their real request-building and response-parsing.

**It immediately caught a live bug.** The awards `category`/`nomination`/`ranking`/`step` sub-routers still declared `/api/club/:clubId<\d+>/…` base URLs from before the slug migration, so *every awards write 404s in production* — the frontend calls `/api/club/${clubSlug}/awards/…`. The old tests hid it by requesting `/api/club/1/…` to satisfy the numeric constraint. Fixed to `:clubSlug` in the four files; that's the only production change in this push besides the test infrastructure.

**Numbers.** Backend statement coverage 63% → 95% (repositories 90%, routes 93%, utils 96%); overall thresholds 73/65/68/74 → 80/69/76/81. 1336 tests, all green. The integration project adds ~45s of container + migration startup, paid once per run; the 230 integration tests themselves take ~18s.

Two things worth flagging:

- **`npm test` now needs Docker.** Added `npm run test:unit` (client + pure backend units, no Docker) for when you don't want it. CI is fine — `ubuntu-latest` ships a daemon. Netlify is unaffected; its build command never ran the tests.
- **One wart pinned rather than fixed:** `og-image.ts` takes its rating count from the raw query rows, and `getReviewsByWorkId` left-joins `review`, so an unscored work renders "from 1 rating" with zero ratings. The test asserts the current behaviour with a comment; counting `scores` instead of `reviews` is the one-line fix, and I left it out to keep the production diff to the awards 404.

`.claude/rules/testing.md` documents the layout, the helpers, and the gotchas (CockroachDB v25+ is required — the `ArbitraryClubLists` migration fails from scratch on v24.1).

---

netlify/edge-functions/shared-review.ts
```diff
@@ -42,11 +42,7 @@ function isCrawler(userAgent: string): boolean {
   return CRAWLER_USER_AGENTS.some((bot) => ua.includes(bot.toLowerCase()));
 }
 
-function generateOGImageUrl(
```
cole-adams, line 45: These look like formatting changes. Were the not caught by the formatting PR check? If not, can you create a follow-on PR that fixes that problem

netlify/functions/tests/helpers/database.ts
```diff
@@ -0,0 +1,53 @@
...
+ * never trips a foreign key (CockroachDB checks constraints per statement).
+ *
+ * `user`, `account`, `session` and `verification` are deliberately absent:
+ * signing a user up costs two bcrypt hashes, so the auth fixtures create their
+ * users once per file and reuse them (see `helpers/auth.ts`). Nothing a test
+ * asserts on is user-global — club membership, lists and reviews all hang off
+ * `club`, which is wiped — so keeping the accounts around is invisible to
+ * tests and saves the suite tens of seconds.
+ */
+const DOMAIN_TABLES = [
```
cole-adams, line 21: Is there a way we can define the type of this array to make sure its a complete array of the keys of the database, then filter out below the tables we don't want to clear? That way when we add a new table, the build will fail until its added here as well?

netlify/functions/tests/helpers/factories.ts
```diff
@@ -0,0 +1,281 @@
...
+import { ClubType, WorkListSystemType, WorkType } from "../../../../lib/types/generated/db";
+import { TMDBMovieData } from "../../../../lib/types/movie";
+import { db } from "../../utils/database";
+import { insertMovieDetails } from "../../utils/movieDetailsUpdater";
+import { tmdbMovie } from "../fixtures/external";
+
+/**
+ * Arrange-phase seeding for the integration suite.
+ *
+ * These write rows directly rather than calling the repositories, so a test's
```
cole-adams: I'm sorta on the fence about this statement. I understand the sentiment, but if we ever change how some of these operations work, we risk this file drifting. Why don't we just use the normal methods for setting things up? Then we would catch problems in tests and as part of fixture setups?

netlify/functions/tests/auth.test.ts
```diff
@@ -0,0 +1,163 @@
...
+}
+
+describe("POST /api/auth/sign-up/email", () => {
+  it("creates the user and sends a verification email", async () => {
+    const email = freshEmail();
+
+    const res = await signUp(email);
+
+    expect(res.statusCode).toBe(200);
+    const user = await db
```
cole-adams: Any direct database reads are a code smell. They are testing implementation details rather than results. Here we just need to test that a user was sent an email

netlify/functions/tests/auth.test.ts
```diff
@@ -0,0 +1,163 @@
...
+    expect(user).toEqual({ name: "New User", emailVerified: false });
+    expect(requestsTo("api.resend.com")).toHaveLength(1);
+  });
+
+  it("stores a hashed password, never the password itself", async () => {
+    const email = freshEmail();
+
+    await signUp(email);
+
+    const account = await db
```
cole-adams: Again, not sure I see value in this test reading from the db

netlify/functions/tests/auth.test.ts
```diff
@@ -0,0 +1,163 @@
...
+      .where("user.email", "=", email)
+      .executeTakeFirstOrThrow();
+    expect(account.password).not.toBe(PASSWORD);
+    expect(account.password).toMatch(/^\$2[aby]\$/);
+  });
+
+  it("answers a repeat sign-up with a decoy instead of creating a second user", async () => {
+    const email = freshEmail();
+    const first = await signUp(email);
+    const original = await db
```
cole-adams: No direct reading, use services

netlify/functions/tests/awards.test.ts
```diff
@@ -0,0 +1,541 @@
...
+
+function awardsData(overrides: Partial<AwardsData> = {}): AwardsData {
+  return { step: AwardsStep.Nominations, awards: [], ...overrides };
+}
+
+/**
+ * Read the stored blob back through the production schema, so a write that
+ * corrupts its shape fails the test rather than being silently reinterpreted.
+ */
+async function storedAwards(clubId: string, year = YEAR): Promise<AwardsData> {
```
cole-adams: Use a service method

netlify/functions/tests/club.test.ts
```diff
@@ -0,0 +1,480 @@
...
+
+    const res = await api.post<{ clubId: string; slug: string }>("/api/club", {
+      body: { name: "The Late Night Crew", members: [] },
+      as: alice,
+    });
+
+    expect(res.statusCode).toBe(200);
+    expect(res.body.slug).toBe("the-late-night-crew");
+
+    const club = await db
```
cole-adams: no direct db calls in tests

cole-adams (commented):
I haven't gone through everything, but I'm seeing a lot of direct db calls in tests. Those should be avoided at all costs, and each usage should be justified. The point of the integration-style test is we should be performing and then asserting results from the same interfaces a client would have access to

brian-norman:
All addressed in 9169b14 — you were right, and following the principle through turned out to strengthen a lot of the tests rather than just relocate them.

**Direct database calls are gone.** Every assertion now reads a write back through the endpoint a client would use: the club preview for a rename, `GET /members` for a profile change, `GET /awards/:year` for an awards write, the list payload for a metadata refresh, `GET /reviews/:workId/scores` for a score. Several got better for it — "renames the club" now proves the rename is visible, not just that a column changed, and "moves the club to the new url" checks the old slug 404s.

**And so is the fixture-drift risk you flagged.** I took your side of the fence: `helpers/factories.ts` now creates clubs, lists, works, scores and comments by posting to the real routes. So `createClub` exercises the same slug generation, default-list creation and settings seeding a real club gets, and a bug in a write path fails the test instead of being papered over.

**`signIn()` is now the whole flow.** It signs up, then **follows the link out of the verification email** — MSW captures Resend's payload, the test pulls the `verify-email` URL out of the HTML and GETs it — then signs in. No more flipping `emailVerified` behind BetterAuth's back. `restoreFixtureUsers` undoes profile edits through `PUT /member/name` and `DELETE /member/avatar` for the same reason.

**Two direct writes survive**, both in `helpers/factories.ts` with a comment saying why:
- `expireInvite()` — `POST /invite` always issues a 24-hour token and nothing can shorten it, so the expiry branches are unreachable through the API.
- `createAwardsYear()` — every awards route runs through `validYear`, which 404s unless the row exists, and no endpoint creates one. Worth a follow-up route.

`scheduled-db-cleanup.test.ts` issues `CREATE DATABASE` for a different reason: its subject operates on databases, not rows. Assertions there go through `DatabaseCleanupRepository.listDatabases()`.

To keep it that way structurally, **`helpers/database.ts` no longer exports the Kysely handle** — it exports `resetDatabase()` and nothing else. A test can't reach the database even if it wants to.

**Your exhaustive-table idea:** done, and it works nicely.

```ts
type DomainTable = Exclude<keyof DB, (typeof PRESERVED_TABLES)[number]>;
const DOMAIN_TABLES = [...] as const satisfies readonly DomainTable[];

type AssertEveryTableHandled<T extends never> = T;
type _EveryTableHandled = AssertEveryTableHandled<
  Exclude<DomainTable, (typeof DOMAIN_TABLES)[number]>
>;
```

I checked it fails the way you'd want — removing `book_details` gives `Type '"book_details"' does not satisfy the constraint 'never'`, naming the offender.

**`shared-review.ts`:** correct, the format check missed it. `format:check` is `oxfmt --check src migrations netlify/functions lib scripts` — `netlify/edge-functions` and `netlify/plugins` aren't in the glob, and `lint` has the same gap. My oxfmt run reformatted them; I've reverted that so this PR's diff stays focused. Say the word and I'll open the follow-on PR widening both globs and committing the resulting reformat.

**Also dropped `utils/tests/validation.test.ts`** — it mocked the repositories to test middleware internals, which is the pattern you're pushing back on. `validClubSlug` and `validListId` are both covered through the routes (unknown slug 404s, a list belonging to another club 404s).

241 integration tests, 1307 total, all green; backend coverage held at 95% statements.

---

brian-norman:
Superseded by a 16-PR stack — same content, reviewable in pieces. Stack #467, all rebased onto latest `main` (`ac2d175`).

**Production fixes (independent of the test work):**
| | |
|---|---|
| #451 | Awards sub-routes still matching `:clubId<\d+>` instead of `:clubSlug` — these four endpoints were returning 404 |
| #452 | `watchUntil` TDZ crash when the predicate already holds; `ListItems` passing the wrong props to `EmptyState` |
| #453 | Accessible names on the icon-only comment actions; export `toMovieDataSummary` |

**Test infrastructure:**
| | |
|---|---|
| #454 | pinia 3, @testing-library/vue 8, jsdom 27, @testcontainers/cockroachdb |
| #455 | Vitest `client` + `server` projects (replacing `root: "src/"`), strict MSW, render-helper and mock updates |

**Backend tests:**
| | |
|---|---|
| #456 | 305 unit tests for `lib/` and `netlify/functions/utils/` |
| #457 | Integration harness (testcontainers CockroachDB) + auth/club/member/members handler suites |
| #458 | Remaining handler suites: awards, lists, reviews, invites, settings, OG image, scheduled functions |

**Frontend tests:**
| | |
|---|---|
| #459 | Service layer + auth store |
| #460 | Shared components and composables |
| #461 | auth, clubs, settings, profile |
| #462 | awards |
| #463 | watch-list |
| #464 | reviews |
| #465 | statistics |

**Finally:**
| | |
|---|---|
| #466 | Coverage thresholds (82/70/77/83) + rewritten `.claude/rules/testing.md` |

### Notes on the split

- Every test file is byte-identical to what was on this branch. Verified with `git diff` against this head: the only residual differences are the deferred coverage config and the oxfmt reflow churn in `netlify/plugins/` and `netlify/edge-functions/`, which was dropped as unrelated noise (`main` is format-clean without it).
- Coverage thresholds land **last**, on purpose. Setting them before the tests exist would put every intermediate PR's CI in the red.
- Each PR was verified green independently: `type-check`, `lint`, `format:check` and the unit projects pass at every layer of the stack (1066 unit tests at the tip). The integration project needs Docker, which wasn't available locally — it ran green in this PR's own CI.
- The awards route bug in #451 is a genuine production defect that the integration tests in #458 surfaced; it's at the bottom of the stack so it can merge first.

PR #384, opened 2026-06-16
Support books for AI discussion questions

brian-norman:
## What

The AI discussion-questions feature was hardcoded for movies. This teaches it about book works, reusing the existing book data model end-to-end — **no schema migration needed** (`WorkType.book`, `book_details`, `book_authors` already exist).

A book club viewing a book review now gets book-appropriate prompts (resolved server-side from the work's title, author(s), and first-publish year), and the "couldn't recognize this ___" fallback reads naturally for books.

## How

Three backend seams + the UI string:

- **`gemini.ts`** — replaced the `(title, releaseYear)` signature with a discriminated union `DiscussionWork = MovieWork | BookWork`. The movie prompt is **byte-for-byte unchanged**; the book variant mirrors its structure (depth ordering, debate framing, succinctness, "return 0 if unrecognized" guard) with book-specific language and a `"Title" by Author (Year)` label.
- **`WorkRepository.getDiscussionContext`** — left-joins `book_details` (`first_publish_year`) and an `authors_agg` CTE (array-aggregating `book_authors`, same pattern `bookProvider` already uses), and selects `work.type` so the handler can branch. The two detail joins are safe because TMDB and OpenLibrary id namespaces are disjoint; `work.type` is the authoritative tiebreaker.
- **`reviews.ts`** — maps the resolved row to the correct `DiscussionWork` variant. Title/metadata stay server-resolved, so the prompt still can't be poisoned by client input.
- **`DiscussionQuestions.vue` / `WorkDetailsContent.vue`** — the one movie-specific UI string now interpolates a `mediaNoun` prop, derived from the same `bookData`/`movieData` split the component already uses.

## Notes

- The discriminated union makes the mapping type-safe: TS forces book fields when `type: "book"`, so a book can't fall through to a movie prompt.
- Requires `GEMINI_API_KEY` (Netlify-console secret), same as the existing movie path — unchanged.

## Testing

- `npm run type-check` — clean
- `npm run lint` — clean
- `npm test` — 106 passed (10 files)

### DIFF
--- a/netlify/functions/repositories/WorkRepository.ts
+++ b/netlify/functions/repositories/WorkRepository.ts
@@ -72,16 +72,36 @@ class WorkRepository {
   }
 
   async getDiscussionContext(clubId: string, workId: string) {
+    // A work is either a movie or a book, so only one detail join matches.
+    // External-id namespaces differ (TMDB ids vs OpenLibrary keys), so the
+    // two detail tables can't cross-match; `work.type` disambiguates anyway.
     return db
+      .with("authors_agg", (qb) =>
+        qb
+          .selectFrom("book_authors")
+          .select([
+            "external_id",
+            db.fn.agg<string[]>("array_agg", ["author_name"]).as("authors"),
+          ])
+          .groupBy("external_id"),
+      )
       .selectFrom("work")
       .leftJoin(
         "movie_details",
         "movie_details.external_id",
         "work.external_id",
       )
+      .leftJoin("book_details", "book_details.external_id", "work.external_id")
+      .leftJoin("authors_agg", "authors_agg.external_id", "work.external_id")
       .where("work.id", "=", workId)
       .where("work.club_id", "=", clubId)
-      .select(["work.title", "movie_details.release_date"])
+      .select([
+        "work.title",
+        "work.type",
+        "movie_details.release_date",
+        "book_details.first_publish_year",
+        "authors_agg.authors",
+      ])
       .executeTakeFirst();
   }
 
--- a/netlify/functions/repositories/WorkRepository.ts
+++ b/netlify/functions/repositories/WorkRepository.ts
@@ -71,37 +71,12 @@ class WorkRepository {
     return insertedWork;
   }
 
-  async getDiscussionContext(clubId: string, workId: string) {
-    // A work is either a movie or a book, so only one detail join matches.
-    // External-id namespaces differ (TMDB ids vs OpenLibrary keys), so the
-    // two detail tables can't cross-match; `work.type` disambiguates anyway.
+  async getById(clubId: string, workId: string) {
     return db
-      .with("authors_agg", (qb) =>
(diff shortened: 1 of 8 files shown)


netlify/functions/repositories/WorkRepository.ts
```diff
@@ -72,16 +72,36 @@ class WorkRepository {
   }
 
   async getDiscussionContext(clubId: string, workId: string) {
+    // A work is either a movie or a book, so only one detail join matches.
```
cole-adams: We should bake the type-specific parts of this query into the work providers. I don't want this query to grow as we add more types. This is the only query in this repo to reference the types, so thinking we can remove it

netlify/functions/utils/gemini.ts
cole-adams, line 1: I'm not a fan that the discussion feature is so coupled with a generic "gemini" util. Why don't we pull out the prompt into the work provider, then this can stay super generic, just the LLM interface

src/features/reviews/components/WorkDetailsContent.vue
```diff
@@ -485,6 +487,11 @@ const cancelDateEdit = () => {
 
 const movieData = computed(() => asMovie(props.movie.original.externalData));
 const bookData = computed(() => asBook(props.movie.original.externalData));
+// Drives book/movie wording in child components (e.g. the discussion-questions
+// "couldn't recognize this ___" message). Mirrors the movieData/bookData split.
+const mediaNoun = computed<"movie" | "book">(() =>
+  isDefined(bookData.value) ? "book" : "movie",
```
cole-adams: We have a frontend service for this. Use that instead

cole-adams (changes requested):

netlify/functions/repositories/WorkRepository.ts
brian-norman: Done in 22503fb — `getDiscussionContext` is gone. The repo now has only a generic `getById` (title, type, external_id), and each provider queries its own detail tables inside its new `getDiscussionPrompt` method.

netlify/functions/utils/gemini.ts
brian-norman: Done in 22503fb — the prompts now live in `movieProvider`/`bookProvider` via a new `MediaProvider.getDiscussionPrompt`. `gemini.ts` is reduced to a generic `generateJson(prompt, responseSchema, zod schema)` with no knowledge of the discussion feature or work types.

src/features/reviews/components/WorkDetailsContent.vue
brian-norman: Done in 22503fb — now uses the club from `useClub` (already loaded in this component) with `clubTypeConfig(club.type).noun` from the club-type registry, same as ReviewView does, instead of sniffing `externalData`.

cole-adams (approved):

PR #386, opened 2026-06-16
Add character names to movie cast

brian-norman:
## What

Adds the **character each actor plays** alongside their name in a movie's cast, persisted to the database.

TMDB's credits response already includes a `character` per cast member (the shared `TMDBCastMember` type even declares it), but it was being dropped before persistence. This wires it all the way through.

## Changes

**Database**
- New migration `20260615_AddMovieActorCharacterNames.ts`: adds a nullable `character_name` column to `movie_actors` and backfills existing rows from TMDB credits (batched 40/req with rate-limit delays), mirroring `20260315_AddPersonProfilePaths.ts`.
- Regenerated `lib/types/generated/db.ts`.

**Backend write path**
- `movieDetailsUpdater.ts`: populate `character_name` in both `insertMovieDetails` (new movies) and `updateMovieDetails` (the daily stale-refresh sweep).

**Backend read path**
- `movieProvider.ts`: include `character` in the `actors_agg` JSON aggregate + its `sql<...>` row type.
- `tmdb.ts`: include `character` in the awards-only `getDetailedMovie` mapper.
- `lib/types/movie.ts`: extend `DetailedMovieData.actors`.

**Frontend**
- `CastList.vue` / `CastModal.vue`: render the character beneath the actor's name. Hidden via `hasValue()` when TMDB returns an empty string (uncredited / "Self" roles).

## Testing

- `npm run type-check` — clean
- `npm run lint` — clean
- `npx vitest run` on cast + filterWorks specs — 18 passed (added a character-rendering assertion to `CastList.spec.ts`)

## Notes for review

- The migration runs a **TMDB backfill** inside `up()`, so it needs `TMDB_API_KEY` in the env (same constraint as the existing `20260315` migration). Because it changes `migrations/`, the deploy preview spawns a dedicated per-PR database and runs the migration there — i.e. the backfill is exercised against a real snapshot in the preview rather than shared `dev`.
- `character_name` is nullable; rows are filled going forward on write and historically by the backfill.


migrations/schema/20260615_AddMovieActorCharacterNames.ts
```diff
@@ -0,0 +1,132 @@
...
+  return response.data;
+}
+
+interface MovieRow {
+  external_id: string;
+  title: string | null;
+}
+
+export async function up(db: Kysely<unknown>) {
+  await sql`ALTER TABLE movie_actors ADD COLUMN IF NOT EXISTS character_name VARCHAR`.execute(
```
cole-adams: I would much prefer all the raw SQL queries in this file be converted to use kysley. If you need, there's a .withTables function that you can use to get the right types into the db object

netlify/functions/utils/providers/movieProvider.ts
```diff
@@ -66,8 +66,12 @@ function detailsQuery(externalIds: string[]) {
         .select([
           "external_id",
           sql<
-            { name: string; profilePath: string | null }[]
-          >`json_agg(json_build_object('name', actor_name, 'profilePath', profile_path) ORDER BY cast_order)`.as(
+            {
+              name: string;
+              character: string | null;
+              profilePath: string | null;
+            }[]
```
cole-adams: I know its existing, but can we use kysely here too?

cole-adams (changes requested):

migrations/schema/20260615_AddMovieActorCharacterNames.ts
brian-norman: Done in 34ba5cc — the ALTERs, SELECT, and bulk UPDATE are now composed with the query builder (using `db.withTables` for typing). Two notes: the VALUES row constructors stay as a small `sql` fragment since Kysely 0.27 has no builder for VALUES lists (this is the pattern from the official Kysely docs), and the explicit `::VARCHAR`/`::INT8` casts are kept because CockroachDB cannot infer placeholder types in a bare VALUES list. The `IF NOT EXISTS` idempotency is preserved via a `db.introspection` guard since `alterTable().addColumn()` has no `ifNotExists()`. Re-verified end-to-end on a freshly spawned dev DB: full backfill (613 movies, 0 errors), `down()`, and re-running `up()` with the column already present.

netlify/functions/utils/providers/movieProvider.ts
brian-norman: Done in 34ba5cc — both aggregates (directors too) now use `eb.fn.jsonAgg()` + `jsonBuildObject()` from `kysely/helpers/postgres`, with `.orderBy("cast_order")` on the aggregate replacing the raw `ORDER BY`. The row types are now derived from the column types instead of asserted via `sql<...>`. Verified the compiled query against a spawned dev DB post-migration.

cole-adams (approved):

PR #392, opened 2026-06-29
Improve review score entry: discoverable /10 scale + roomy drawer input

brian-norman:
## Summary

New users don't intuitively know that review scores are out of 10, or that decimal precision is allowed — and the inline score input was uncomfortably small, especially on a gallery poster chip. This PR makes the scale discoverable and moves score entry somewhere with room to type.

## Changes

**Discoverable scale & decimals** (`ReviewScore.vue`)
- `type="number"` + `inputmode="decimal"` so the field gets numeric controls and a numeric/decimal keyboard on mobile
- `min="0"`, `max="10"`, `step="any"` to reflect the valid range (decimals allowed), with the desktop spinner arrows hidden for a clean look
- A `/10` suffix and an `8.5` placeholder so the out-of-10 scale and decimal support are visible while editing
- Submit on blur (Enter blurs to submit), skipping the mutation when the value is unchanged

**Gallery scores entered via the drawer, not the poster chip**
- Tapping a gallery poster's score affordance no longer opens a cramped inline input. It opens the work-details drawer — bottom sheet on mobile, side panel on desktop — with the score field already open and focused.
- `ReviewScore` gains `openInDrawer` (defer to the drawer) and `autoFocus` (open + focus on mount). `GalleryView` provides a `requestScoreEntry` handler through a typed `InjectionKey` and opens the drawer with a `focusScoreEntry` flag threaded down to the current user's score cell in both score grids.
- The `/10` hint is omitted on the poster chips (no room); it remains on the table cells and the drawer input.
- Table view is unchanged — inline editing stays, since those cells have space.

All rating surfaces (table, gallery cards, detail drawer) render through `ReviewScore.vue`, so these changes apply consistently across mobile and desktop.

## Notes for review
- **iOS soft keyboard:** the drawer input is focused programmatically in `onMounted` (after the sheet animates in), so iOS Safari may show the field focused but require one extra tap to raise the keyboard. Worth an eyeball on a real device.
- No automated coverage yet for the new gallery → drawer path; existing table-view inline-entry test still passes.


cole-adams (changes requested):
This looks horrible on desktop:

Image

brian-norman:
Good catch — that was the native number-input stepper arrows crowding the small score box. Pushed a fix (a8e0d9a) that hides the spinners across browsers (`appearance-none` on the WebKit/Blink spin pseudo-elements + `appearance:textfield` for Firefox) via Tailwind arbitrary variants. `type="number"`/`inputmode="decimal"` are unchanged, so mobile still gets the numeric keyboard.

cole-adams (approved):

PR #393, opened 2026-06-30
Remove scroll bar from filter pills

brian-norman:
## Summary

Gets rid of the scroll bar on the filter pills in `SearchFilterBar.vue`.

Previously the pill row used `flex-nowrap` + `overflow-x-auto` on mobile (so pills scrolled horizontally) and only wrapped on desktop (`md:flex-wrap`). A `scrollbar-hide` CSS hack was used to mask the resulting scroll bar.

Now the pills simply wrap on all viewports (`flex-wrap justify-center`), so there's no overflow container to render a scroll bar at all. The now-unused `scrollbar-hide` style block has been removed.

## Changes

- `src/common/components/SearchFilterBar.vue`: replace `flex-nowrap gap-2 overflow-x-auto md:flex-wrap md:justify-center` + `scrollbar-hide` with `flex-wrap justify-center gap-2`, and drop the unused `scrollbar-hide` scoped CSS.

---
_Generated by [Claude Code](https://claude.ai/code/session_01NgusquQWdjZM7MRzmcjA3b)_

### DIFF
--- a/src/common/components/SearchFilterBar.vue
+++ b/src/common/components/SearchFilterBar.vue
@@ -26,7 +26,7 @@
     <!-- Available filter options as pills -->
     <PopoverGroup
       as="div"
-      class="scrollbar-hide relative mt-2 flex w-full flex-nowrap gap-2 overflow-x-auto md:flex-wrap md:justify-center"
+      class="relative mt-2 flex w-full flex-wrap justify-center gap-2"
     >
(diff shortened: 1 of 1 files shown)


cole-adams:
Maybe this is just me, but I think I like the way it is now better. I've looked at a few different mobile apps with filtering, and they all use this side-scroll pill for different filter fields. It seems intuitive to me?

PR #397, opened 2026-07-03
Fix awards results showing blank rankings after a member renames

brian-norman:
## Problem

On a completed awards session, a member's ranking column rendered **blank** while other members showed fine.

**Root cause:** rankings (`ranking`) and nominations (`nominatedBy`) were keyed by the submitter's **display name at submit time**. When a member later renamed themselves, the stored key no longer matched their current name, so the results page's `nomination.ranking[member.name]` lookup returned `undefined`. Because display names are user-editable (`PUT /api/member/name`), this is a latent bug for *any* member who ever renames — not a one-off.

A read-only scan of all `awards_temp` rows confirmed exactly one orphaned key across the whole DB (`brian`, renamed to `Brian`), but the fix addresses the class of bug.

## Fix

Key rankings and nominations by **stable user ID** instead of display name.

- **Writes** (`src/service/useAwards.ts`): submit `user.id` instead of `user.name` for nominations and rankings, including optimistic cache updates.
- **Reads / lookups**: `AwardResult.vue`, `AwardRanking.vue`, `NominationsView.vue`, `YearView.vue` now key off `member.id` / `user.id`. The one place a stored key is *displayed* (the `nominatedBy` avatars in `AwardRanking.vue`) resolves `id → name` via the members list.
- **Backend** (`ranking.ts`, `nomination.ts`): pure pass-through, so only clarifying comments that these fields now carry user IDs.
- Also folds in two related results-page fixes already in progress: results-card grid spacing, and renaming the mis-declared `clubId` prop to `clubSlug` (the router only supplies `clubSlug`).

## Migration

`migrations/schema/20260706_RekeyAwardsToUserIds.ts` rekeys existing `awards_temp` rows from names to user IDs. It's a regular **Kysely schema migration** (originally a `migrate:data` script; converted per review feedback), so it runs automatically through the deploy pipeline — including on this PR's own isolated `pr_397` preview database, which the `preview-database` plugin now spawns because the PR touches `migrations/schema/`.

Properties:

- **Idempotent and lossless**: keys already equal to a valid member ID are left alone; unresolvable keys are logged but never dropped.
- Auto-maps every current member by name, with a single manual override for the one already-renamed member (`brian` → Brian's user ID).
- Unresolved keys don't fail the migration (it runs unattended on deploy) — they were already unreadable under name-keying and stay inert under ID-keying.
- Includes a best-effort `down()` for dev (`migrate:down`) that maps IDs back to current display names.
- The awards JSON shape is validated against a **frozen local copy** of the zod schema rather than importing `lib/types/awards.ts`, so future type changes can't rewrite what this migration validates.

## Verification

- `npm run type-check` ✅, `npm run lint` ✅, `npm test` ✅ (106 passing)
- Ran the migrator end-to-end against a **freshly spawned dev database** (never shared `dev`): all pending schema migrations applied cleanly, the rekey rewrote all **8 rows with 0 unresolved keys** (`brian` correctly resolved to Brian's ID, verified by inspecting the stored JSON), and `down` → `up` round-trips cleanly. The throwaway DB was deleted afterward.
- **Browser check (local):** on a completed session, every member's ranking column now renders — including the previously-blank column for the renamed member (screenshot below).

pr-397-awards-results

> Note: the awards feature has no automated test coverage, so behavior was verified via the type system, the spawned-DB migration run, and the local browser check above.


cole-adams (changes requested):
I see the bubbles again, but I don't see the ranking scores:

Image

brian-norman:
> I see the bubbles again, but I don't see the ranking scores:
> Image

There wasn't a preview DB for me to run the data migration against for this PR since there wasn't a schema migration

cole-adams:
> > I see the bubbles again, but I don't see the ranking scores:
> > Image
> 
> There wasn't a preview DB for me to run the data migration against for this PR since there wasn't a schema migration

Oh I see. Can we just put the migration in the kysely migrations? Honestly I don't think we need the data migration flow anymore. I think we can just do everything we need todo in the kysely migrations

brian-norman:
Done in c4722c5 — the rekey is now a regular Kysely schema migration (`migrations/schema/20260706_RekeyAwardsToUserIds.ts`) and the data migration is deleted. Since the PR now touches `migrations/schema/`, the preview-database plugin will spawn an isolated `pr_397` DB and the deploy pipeline runs the migration there automatically, so you can verify it on the deploy preview directly.

Verified locally against a freshly spawned dev DB: all 8 `awards_temp` rows rekeyed with 0 unresolved keys (including the `brian` → Brian override), and it round-trips cleanly through `migrate:down`/up. Two notes from the conversion: unresolved keys log a warning instead of failing (it runs unattended on deploy), and the awards JSON is validated against a frozen local copy of the zod schema so future type changes can't affect what this migration does.

Agreed on retiring the data migration flow — the remaining `migrations/data/` scripts are all one-shots that have already run. I'd do that removal (`migrate:data` script + `dataMigrator` + old files) as a small follow-up PR rather than folding it in here.

cole-adams (approved):

PR #400, opened 2026-07-05
Motion system phase 2: micro-interactions

brian-norman:
## Summary

Phase 2 of 3 (stacked on #399 — merge that first). Adds motion to the highest-traffic interactions, all on the Phase 1 tokens:

- **Button press feedback**: `VBtn` scales to 0.97 on `:active` (fast/standard tokens); `MenuCard` lifts 2px on hover and scales to 0.98 on press.
- **Score editing** (the app's most repeated action): `ReviewScore` now swaps between score / add / input states through a 150ms out-in scale-fade. Input focus moved from `nextTick` to the transition's `@after-enter` hook — with `mode="out-in"` the input only mounts after the old element finishes leaving, so `nextTick` fires too early.
- **Search results stagger**: results in the add-work modal enter with a fade + 8px rise, staggered 25ms per card (capped at 8 cards / +200ms) via a per-card `--stagger` CSS variable — no JS animation hooks. Removal is deliberately instant so typing never lags the filter. `appear` is required: both grids mount together with their `v-if` parent, and Vue only animates insertions into an already-rendered TransitionGroup.
- Skeletons were audited and already consistent (`bg-lowBackground` + `animate-pulse`), so no change there.

## Test infrastructure note

`src/tests/utils.ts` now merges per-test `global.stubs` overrides instead of clobbering them. Vue Test Utils stubs `<Transition>` **process-wide by default** (via Vue's `transformVNodeArgs`), which silently swallows all transition hooks in every jsdom test — the score test disables that stub to exercise the real after-enter focus path, and awaits the now-asynchronous focus/removal.

## Verification

- Playwright (Chromium): clicking a score opens the input with focus landing correctly after the swap; blur closes it. Add-work modal cards enter with observed computed delays `0s → 0.2s` in 25ms steps; junk queries clear all cards within 120ms.
- Accidental bonus check: with `prefers-reduced-motion` emulated, the stagger delays collapse to 0s via the Phase 1 reduced-motion layer.
- `vue-tsc`, `eslint --max-warnings 0` green; `npm test` 116/117 (the 1 failure is the pre-existing `statsComputers` tie-ordering issue on main, unrelated).

### DIFF
--- a/src/common/components/VBtn.vue
+++ b/src/common/components/VBtn.vue
@@ -1,9 +1,9 @@
 <template>
   <button
-    class="rounded-md text-center text-base font-bold tracking-wide text-text filter duration-fast ease-standard"
+    class="rounded-md text-center text-base font-bold tracking-wide text-text transition duration-fast ease-standard"
     :class="{
       'bg-gray-600': disabled,
-      'cursor-pointer bg-primary hover:brightness-110 active:brightness-105':
+      'cursor-pointer bg-primary hover:brightness-110 active:scale-[0.97] active:brightness-105':
         !disabled,
     }"
     :disabled="disabled"
--- a/src/common/components/WorkSearchPrompt.vue
+++ b/src/common/components/WorkSearchPrompt.vue
@@ -11,20 +11,24 @@
         <h5 class="text-left font-bold">
           {{ defaultListTitle }}
         </h5>
-        <div
+        <TransitionGroup
+          tag="div"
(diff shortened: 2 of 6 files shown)


cole-adams:
Looks like this needs a rebase for the deploy preview

brian-norman:
> Looks like this needs a rebase for the deploy preview

done

cole-adams (approved):

PR #406, opened 2026-07-07
Live-sync scores after submitting via a per-movie poll

brian-norman:
## Problem

Club members usually score a movie/book at the same time. After everyone submits, each person has to hard-refresh the page a few times before the others' scores show up.

That's because the reviews list query (`["list", <slug>, "reviews"]`) never refetches on its own — `refetchOnWindowFocus` is off and `main.ts` has a custom `refetchOnMount` that stops refetching after 2 fetches per session. A member's new score only reaches your client when *you* perform a mutation (which invalidates that key).

## Fix

Add a lightweight per-work scores endpoint and, after a member submits/edits their own score, poll it for a short window so everyone else's scores stream in without a manual refresh.

- **`GET /api/club/:clubSlug/reviews/:workId/scores`** (`secured`) returns *only* the `scores` map for one work (per-member entries plus the synthetic `average`) — the exact shape the frontend already stores per list item. It reuses the existing single-work query `ReviewRepository.getReviewsByWorkId` and filters to current members, so posters/metadata are never re-fetched.
- After `useReviewWork` / `useUpdateReviewScore` succeed, a module-scoped poller hits that endpoint **10 times at 0.5s intervals** and merges each response into the cached reviews list, replacing only the matching work's `scores`. A token-keyed map prevents overlapping loops when the same work is re-scored during the window.
- Existing optimistic update + `onSettled` invalidation are kept: the invalidation gives the submitter their own authoritative row immediately; the poll catches everyone else over the following ~5s.

Polling only fires right after a submission — there's no global `refetchInterval`, so idle/page-load traffic is unchanged.

## Files

- `netlify/functions/club/reviews.ts` — new `/:workId/scores` route + `buildWorkScores` helper
- `lib/types/lists.ts` — shared `ReviewScores` type
- `src/service/useReviews.ts` — `startScorePoll` + `onSuccess` wiring on both score mutations
- `src/features/reviews/components/ReviewScore.vue` — thread `workId` into the edit call
- `src/mocks/handlers.ts` — MSW handler for the new endpoint (strict MSW)

## Verification

- `npm run type-check`, `npm run lint`, `npm test` (154 passing) all green.
- Manual two-client check still recommended: with two members on the Reviews view, a score submitted by one appears for the other within ~5s without a refresh.

### DIFF
--- a/netlify/functions/club/reviews.ts
+++ b/netlify/functions/club/reviews.ts
@@ -1,9 +1,11 @@
 import { z } from "zod";
 
-import { hasValue } from "../../../lib/checks/checks.js";
+import { hasValue, isDefined } from "../../../lib/checks/checks.js";
+import { ReviewScores } from "../../../lib/types/lists";
 import ListRepository from "../repositories/ListRepository";
 import ReviewRepository from "../repositories/ReviewRepository";
 import SettingsRepository from "../repositories/SettingsRepository";
+import UserRepository from "../repositories/UserRepository";
 import WorkCommentRepository from "../repositories/WorkCommentRepository";
 import WorkRepository from "../repositories/WorkRepository";
 import SharedReviewService from "../services/SharedReviewService";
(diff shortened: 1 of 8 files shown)


netlify/functions/club/reviews.ts
```diff
@@ -207,6 +209,62 @@ router.post(
...
+
+  const userScores = reviews.reduce<ReviewScores>((acc, review) => {
+    if (
+      hasValue(review.user_id) &&
+      hasValue(review.review_id) &&
+      hasValue(review.score) &&
+      isDefined(review.created_date) &&
+      memberIds.has(review.user_id)
+    ) {
+      acc[review.user_id] = {
```
cole-adams: This is identical to the code in `list.ts`. Can we pull out a helper

brian-norman: Good call — extracted into a shared `buildReviewScores(entries, memberIds)` helper in `netlify/functions/utils/reviewScores.ts`. Both `getReviewList` (list.ts) and `buildWorkScores` (reviews.ts) now delegate to it. The one wrinkle was that the two queries name their date column differently (`time_added` vs `created_date`), so each caller maps its rows into a common `ScoreEntry` shape at the boundary rather than pushing that difference into the helper. Net −52 lines across the two call sites. (55336ce)

cole-adams: Wait that's a great catch on the query. They shouldn't be different. The old query was wrong, it should have been using `created_date` not `time_added`. `created_date` is the timestamp the review was created and `time_added` is the timestamp the work was added to the list.

Please remove the mapping so that they both use the correct `created_date`

cole-adams (changes requested):

netlify/functions/club/reviews.ts
brian-norman: Done — good catch on the semantics. `getReviewList` now selects `review.created_date` directly (the review's own timestamp), so both queries expose the same column and the `time_added` → `created_date` remapping in `buildReviewScores` is gone. `time_added` is still used for the list item's own `createdDate` (when the work was added to the list), which is correct. (a0ac90d)

cole-adams (approved):

PR #407, opened 2026-07-07
Redesign score entry: arc-gauge dial + Score Assist comparisons

brian-norman:
## What

Redesigns score entry around a single reusable panel and adds **Score Assist**, a comparison flow for members who aren't sure what to score.

**Arc-gauge dial** — the score field is a semicircular gauge: drag the handle along the arc (snapping to 0.5 steps) or type a precise score in the center input. Verdict bands (Awful → Masterpiece) color the arc, readout, and label. Typing is validated live: values past two decimals are truncated, out-of-range values clamp with an explanatory pill ("Max is 10 — set to 10.0"), and either correction shakes the input.

**One entry surface everywhere** — the details drawer's footer CTA ("Rate this movie" / "Edit score") expands in place into the score panel, and the reviews table's score cell opens the same panel in an anchored popover. The Save button matches the CTA's size and shape, so the footer keeps one consistent button across states.

**Score Assist** — members who have scored ≥5 works get a "Compare movies you've rated" button. It asks up to five "Which did you like more?" questions — answered by tapping the poster cards directly (hover ring + "Tap the one you liked more" hint) — binary-searching the member's own 0–10 scale. When the session converges, the suggested score is saved immediately and a toast announces it ("We picked 8/10 for Dune"). "Too close to call" adopts the compared work's score; "Skip" swaps in another work.

## How

- The assist algorithm is a pure, unit-tested state machine (`scoreAssistLogic.ts`). Pivots come from the rank-middle third of eligible candidates (so clumped scores still halve the search space), preferring works most similar to the target (shared directors/genres for movies, authors/subjects for books) via a feature-local `Record<WorkType, ...>` registry.
- Eligibility is computed client-side from the cached reviews query — no new endpoints or schema changes.
- Copy derives from `clubTypeConfig(type).noun`, so book clubs read "Which book did you like more?" with no inline club-type branches.
- `scoreScale.ts` is the single source of truth for the 0–10 bounds/step; `useSubmitScore(clubSlug)` collapses the repeated "reviewId ? update : create" mutation.
- `useBackButtonClose` now coordinates stacked overlays: closing the top layer no longer dismisses the overlay beneath it.

## Testing

214 tests pass (algorithm, similarity, and dial-geometry units; component specs for the panel, dock, modals, and assist flow with MSW spies), plus lint and type-check. Verified end-to-end in a real browser against `netlify dev` on desktop and mobile: dial drag/typing with clamp + shake, drawer and table entry points, and the full compare flow through auto-save and toast.


src/common/haptics.ts
```diff
@@ -0,0 +1,36 @@
...
+ * - iOS (18+): Safari never implemented `navigator.vibrate`, but toggling a
+ *   native switch control (`<input type="checkbox" switch>`) via a label
+ *   click fires the system haptic tick, so we synthesize one off-DOM.
+ *
+ * Everywhere else (desktop, older iOS) this is a harmless no-op.
+ */
+
+const TICK_MS = 10;
+
+const clickHiddenSwitch = (): void => {
```
cole-adams: This seems overly hacky. Maybe just remove

src/common/workDisplay.ts
```diff
@@ -69,3 +69,40 @@ export function workSubtitle(
...
+  if (hours === 0) return `${rest}m`;
+  return rest === 0 ? `${hours}h` : `${hours}h ${rest}m`;
+}
+
+/**
+ * A one-line, media-appropriate summary shown under the title in the details
+ * drawers: "2h 35m · Adventure, Science Fiction" for movies, "Frank Herbert ·
+ * 412 pages" for books. Undefined when there is nothing to show.
+ */
+export function workMetaLine(
```
cole-adams: Lets move this function to the club type config

src/features/reviews/components/ScoreAssistFlow.vue
```diff
@@ -0,0 +1,146 @@
...
+  result,
+  answer: answerSession,
+  progressLabel,
+} = useScoreAssist(props.target, props.candidates);
+
+const noun = computed(() => clubTypeConfig(props.clubType).noun);
+
+// The whole card is the answer button; hover/focus affordances (ring + lift)
+// signal that the posters themselves are what you tap.
+const posterButtonClass =
```
cole-adams: Why not just inline this in the two places? We don't do this anywhere else. It's generally an anti-pattern to set a list of classes to a variable in tailwind. If you really need to keep them common, create a new component

src/features/reviews/components/ScoreDial.vue
```diff
@@ -0,0 +1,246 @@
...
+// The draft score stays a string so the input round-trips partial typing
+// ("8.", "") and preserves the exact decimals the user entered; the arc only
+// reacts once the string parses to a valid score.
+const model = defineModel<string>({ required: true });
+
+const emit = defineEmits<{
+  (e: "save"): void;
+}>();
+
+const arcTransition = "transition-all duration-fast ease-standard";
```
cole-adams: No class variables

src/features/reviews/components/WorkDetailsContent.vue
```diff
@@ -508,6 +395,35 @@ const displayYear = computed(() =>
...
 
+// "2h 35m · Adventure, Science Fiction" (movies) / "Frank Herbert · 412 pages"
+// (books), shown in the hero under the title. Runtime and genres live here, so
+// the Details section below only carries what the hero doesn't.
+const metaLine = computed(() =>
+  workMetaLine(props.movie.original.externalData),
+);
+
+const overview = computed(
+  () => movieData.value?.overview ?? bookData.value?.description,
```
cole-adams: Club type config

src/features/reviews/components/WorkDetailsContent.vue
cole-adams, line 1: Outside the scope of this PR, but I would love to see this refactored in such a way that the club config contains the information and components to render this panel for each of the club types. This component has huge risk of growing with club types as it has so much type-specific logic

src/features/reviews/composables/scoreAssistSimilarity.ts
```diff
@@ -0,0 +1,67 @@
...
+}
+
+/**
+ * Feature-local club-type registry (see .claude/rules/code-quality.md):
+ * similarity is Score Assist logic over the lib work-data shapes, so it lives
+ * here keyed by WorkType rather than in the shared CLUB_TYPE_CONFIG. The
+ * exhaustive Record means a new work type won't compile until it defines its
+ * notion of similarity.
+ */
+const SIMILARITY_BY_WORK_TYPE: Record<WorkType, WorkSimilarity> = {
```
cole-adams: Can we include this logic in the club type config? I don't think it just needs to be copy. We can expand all club type logic into subfiles and just register them in the config

src/features/reviews/scoreBands.ts
```diff
@@ -0,0 +1,26 @@
+/**
+ * Verdict bands for the score dial: each score maps to a label and an accent
+ * color used for the arc fill, the score readout, and the clamp notice pill.
+ * Registry-style (like CLUB_TYPE_CONFIG) so new bands are one entry here, not
```
cole-adams, line 4: I think this reads a little too much into the claude rules LOL, but not a bad way to do it

src/features/watch-list/components/ListItemDetailsContent.vue
```diff
@@ -208,4 +217,13 @@ const posterUrl = computed(() =>
 
 // Release year (movies) or first-published year (books), via the shared helper.
 const displayYear = computed(() => workSubtitle(props.movie.externalData));
+
+// "2h 35m · Adventure, Science Fiction" (movies) / "Frank Herbert · 412 pages"
+// (books), shown in the hero under the title. Runtime and genres live here, so
+// the Details section below only carries what the hero doesn't.
+const metaLine = computed(() => workMetaLine(props.movie.externalData));
+
+const overview = computed(
```
cole-adams: Use club type config

src/features/watch-list/components/ListItemDetailsContent.vue
```diff
@@ -208,4 +217,13 @@ const posterUrl = computed(() =>
 
 // Release year (movies) or first-published year (books), via the shared helper.
 const displayYear = computed(() => workSubtitle(props.movie.externalData));
+
+// "2h 35m · Adventure, Science Fiction" (movies) / "Frank Herbert · 412 pages"
+// (books), shown in the hero under the title. Runtime and genres live here, so
+// the Details section below only carries what the hero doesn't.
+const metaLine = computed(() => workMetaLine(props.movie.externalData));
```
cole-adams, line 219: Use club type config

src/service/useReviews.ts
```diff
@@ -186,6 +186,34 @@ export function useUpdateReviewScore(clubSlug: string) {
...
+  return ({
+    workId,
+    reviewId,
+    score,
+  }: {
+    workId: string;
+    reviewId?: string;
+    score: number;
+  }) => {
+    if (isDefined(reviewId) && reviewId !== "") {
```
cole-adams: Helper `hasValue` does this check

cole-adams (changes requested):

src/common/haptics.ts
brian-norman: The hacky part was the iOS-Safari fallback (it clicked a hidden native `<switch>` to borrow a system haptic tick). I ripped that out in `5410b55` — `hapticTick` is now just the standard Vibration API (~5 lines), a clean no-op on Safari/desktop, used only for the dial's detent ticks on Android. Happy to delete it wholesale if you'd still rather not carry the feature, but the remaining code isn't doing anything sketchy anymore.

src/common/workDisplay.ts
brian-norman: Done in `46f5951`. `subtitle`/`metaLine`/`overview` now live on `CLUB_TYPE_CONFIG[type].display`, and `workSubtitle`/`workMetaLine`/`workOverview` are thin helpers that dispatch on `externalData.kind` — no more inline movie-vs-book branch. (Kept the narrowers `asMovie`/`asBook` and `workPosterUrl` in `workDisplay.ts` to avoid a `clubType → workDisplay → clubType` import cycle.)

src/features/reviews/components/ScoreAssistFlow.vue
brian-norman: Inlined into both buttons and dropped the variable (`46f5951`).

src/features/reviews/components/ScoreDial.vue
brian-norman: Inlined the arc transition into both `:class` bindings (`46f5951`).

src/features/reviews/components/WorkDetailsContent.vue
brian-norman: `overview` now reads from the registry via `workOverview(...)` instead of `movieData?.overview ?? bookData?.description` (`46f5951`).

src/features/reviews/components/WorkDetailsContent.vue
brian-norman: Agreed — that's the right long-term shape, and I've left the full "config owns the panel components per club type" refactor as follow-up since it's bigger than this PR. The display + similarity extraction here moves in that direction: per-type logic is now registered on the config rather than sprinkled through the component.

src/features/reviews/composables/scoreAssistSimilarity.ts
brian-norman: Moved it onto the config: each entry now has a `similarity(target, candidate)` and `workSimilarity(type, ...)` just delegates. Deleted `scoreAssistSimilarity.ts` entirely (`46f5951`). Worth noting it only ever touched `lib`/`common` shapes (`DetailedWorkData`, `asMovie`/`asBook`), so putting it in `common/clubType.ts` doesn't invert the `common → feature` dependency the way the statistics `WorkStatsData` case would.

src/features/reviews/scoreBands.ts
brian-norman: Ha, fair 😄 — left the bands as-is. They're a plain data table so the "registry" framing is basically free here; not worth restructuring.

src/features/watch-list/components/ListItemDetailsContent.vue
brian-norman: `overview` here is registry-driven now via `workOverview(...)` (`46f5951`).

src/features/watch-list/components/ListItemDetailsContent.vue
brian-norman: `metaLine`/`subtitle`/`overview` all go through the club-type config helpers now (`46f5951`).

src/service/useReviews.ts
brian-norman: Switched to `hasValue(reviewId)` (`46f5951`).

brian-norman:
Thanks for the review @cole-adams — addressed everything in `46f5951`:

- **Club-type config** — moved `subtitle`/`metaLine`/`overview` and Score Assist `similarity` onto `CLUB_TYPE_CONFIG`; the shared helpers dispatch on `externalData.kind`/`WorkType` and `scoreAssistSimilarity.ts` is gone. Per-type logic is now registered on the config rather than branched inline.
- **No class variables** — inlined the ScoreDial arc-transition and ScoreAssistFlow poster-button classes.
- **`hasValue(reviewId)`** in `useSubmitScore`.
- **Haptics** — the hacky iOS fallback was already removed in `5410b55`; what's left is a plain Vibration-API tick. Flagged inline in case you'd still prefer to drop it entirely.
- Left the "config owns the panel components per club type" rework as a follow-up (noted inline) since it's larger than this PR.

`type-check`, `lint`, and all 218 tests pass.

cole-adams (approved):

PR #413, opened 2026-07-13
Add spotlight review facts with share shortcut

brian-norman:
## Summary

Once **every club member has scored** a work, one "spotlight fact" appears below the club average — in the review drawer and on the shared review page. Most reviews intentionally produce nothing; facts are gated by thresholds so they stay special.

### Fact kinds (ranked, first match wins)

1. **All-time record** — highest/lowest-rated in club history (needs 10+ scored works)
2. **Count milestone** — 10th, 25th, 50th, 100th, … review
3. **Most divisive ever** — widest score spread in club history (spread >= 4, 10+ works)
4. **Best/worst of the year** — needs a 5+ work cohort that year
5. **Director/author record** — best/worst of 3+ works by the same person
6. **Actor milestone** — 5th, 10th, … movie featuring the same actor
7. **First genre** — first movie of a genre (after 10+ works of history)
8. **TMDB hot take** — club average deviates >= 2.5 points from TMDB

### Implementation notes

- `computeReviewFact()` is one pure function computed client-side from the public reviews-list endpoint (the same one the shared statistics page uses), so the drawer and the anonymous share page render the identical fact with no new backend.
- Generators live in a `Record<WorkType, FactGenerator[]>` registry (same pattern as `WORK_STATS_BUILDERS`) — book clubs get author records instead of director/actor/genre facts, and a new work type will not compile without declaring its list.
- The fact card gets gradient-border + glow treatment and carries a share icon button that re-shares the review link (native share on mobile, clipboard on desktop).
- Count-based facts only look at works up to the target, so "your 50th movie" stays true when revisited later; records compare against full history.
- New mdi icons are registered in `src/icons.ts`, with a `FACT_ICONS` registry test in `icons.test.ts` (fact icons reach templates via a computed, invisible to the static scan).

### Testing

- 13 unit tests for the fact engine (gating, ranking, each generator)
- Verified in the running app: drawer + share page both render the fact ("Hot take" fires for *Something* in the cole-test dev club), and the share button copies the link


src/features/reviews/components/WorkDetailsContent.vue
```diff
@@ -376,6 +406,56 @@ const cancelDateEdit = () => {
...
+// for the drawer's opening frame, so the computed stays null until the browser
+// goes idle after mount — the drawer paints and slides in immediately, then
+// the fact card transitions in. The component is keyed by movie.id, so the
+// gate re-arms on every open/work switch.
+//
+// The dismiss animation needs the same protection: the drawer stays mounted
+// until its leave transition finishes, so a still-pending idle callback (or
+// its timeout) can land mid-close — the `dismissing` guard skips the scan
+// then, since a fact appearing in a closing drawer helps no one.
+const { data: allReviews } = useReviewsList(clubId);
```
cole-adams, line 417: Doesn't this kinda defeat the purpose of your new PR sliming down what we need to render reviews? Should be do review fact computation serverside?

src/features/reviews/reviewFacts.ts
```diff
@@ -0,0 +1,616 @@
...
+export function computeReviewFact(
+  reviews: DetailedReviewListItem[],
+  members: readonly { id: string }[],
+  workId: string,
+): ReviewFact | null {
+  if (!hasElements(members)) return null;
+  const target = reviews.find((review) => review.id === workId);
+  if (!isDefined(target)) return null;
+
+  const allScored = members.every((member) => {
```
cole-adams: Why this restriction? A member who isn't reviewing works would prevent facts from showing up. With how the book club is setup, they would never see any facts because you don't review the books

src/features/reviews/reviewFacts.ts
```diff
@@ -0,0 +1,616 @@
...
+ *
+ * Facts only unlock once *every* club member has scored the work — that keeps
+ * them from spoiling scores mid-round and marks the moment a review is
+ * "complete". Most reviews won't produce one; that's by design.
+ */
+export function computeReviewFact(
+  reviews: DetailedReviewListItem[],
+  members: readonly { id: string }[],
+  workId: string,
+): ReviewFact | null {
```
cole-adams: nit: prefer undefined over null

cole-adams (changes requested):

src/features/reviews/components/WorkDetailsContent.vue
brian-norman: Good callout. On the reviews page this doesn't actually add a fetch — the reviews table already loads `/list/reviews` to render, so the fact computation just reuses the cached detailed list. `SharedReviewView` already fetches it too for the public shared page. So there's no extra payload cost today.

You're right that it couples facts to the heavy detailed payload, which cuts against the slimming in #421. Rather than stand up a dedicated server endpoint in this PR, I'd like to leave it client-side for now and revisit once #421 lands: `computeReviewFact` is already a pure function, so moving it behind a small `GET /reviews/:workId/fact` endpoint at that point is a clean lift. Flagging it as a known follow-up.

src/features/reviews/reviewFacts.ts
brian-norman: Good catch — fixed in a5f3bca. Dropped the every-member gate entirely; a fact now shows as soon as the work has a club average (i.e. at least one score). That unblocks book clubs where not everyone reviews each work. Also removed the now-unused `members` argument and the `useMembers` fetch from the drawer, and added a test asserting a fact surfaces when only some members have scored.

src/features/reviews/reviewFacts.ts
brian-norman: Done in a5f3bca — `computeReviewFact` and all the generators now return `undefined`.

cole-adams (approved):

PR #421, opened 2026-07-16
Cut API latency: slim payloads, fewer round trips, smarter caching

brian-norman:
## Why

Loading clubs, reviews, adding movies, and moving movies to reviews all felt slow. Investigation found three compounding causes, measured against production:

1. **Cross-region multiplier**: Netlify Functions run in us-east-1 while CockroachDB is in aws-us-west-2 — every DB query pays ~65-70ms, and hot paths issued 10-15 _sequential_ queries (add-movie: middleware ×4 + lookup/insert chain + a synchronous TMDB fetch + 6 sequential metadata inserts).
2. **Payload weight**: the reviews endpoint shipped **2.67 MB**, 78% of it full cast lists the list view never renders (only the detail drawer and one stats widget need them).
3. **Broken cache persistence**: the query cache persisted through the _synchronous localStorage_ persister — one large club's payload alone approaches the ~5MB quota, so persistence likely failed silently and every session started cold. (The in-memory `fetchedMap` mount-gate that governs refetch-on-mount was already correct and is kept as-is.)

## What

**Payloads** — list/review endpoints now return a summary shape (`WorkDataSummary`): full `actors` objects removed, lightweight `castNames` kept so the `actor:` search filter still works. Full cast ships on demand via `GET /work/:workId/details` (detail drawers) and `GET /reviews/cast` (statistics leaderboards). Largest club's reviews payload: **2.67 MB → 0.67 MB (−75%)**, verified locally.

**Round trips** —

- `POST /:listId/items`: work upsert without pre-lookup, TMDB fetch skipped when `movie_details` is already cached, junction inserts parallelized, and item insertion is one `INSERT..SELECT` that computes position and detects duplicates via the unique constraint (was: 5 sequential queries + unconditional TMDB call).
- `moveItem` validates destination ownership inside its existing transaction query.
- `GET /:clubSlug` reuses the club row `validClubSlug` already fetched; `isUserInClub` is a single `club_member` index lookup.
- New `GET /list/all-items` replaces the client-side one-request-per-list fan-out (each of which paid its own function invocation + middleware queries).
- BetterAuth `session.cookieCache` (5 min) skips the session DB query on most secured requests.

**Auth handler fix (found during verification)** — the Netlify auth adapter copied headers via `Headers.forEach`, which folds repeated `Set-Cookie` headers into a single comma-joined (invalid) header. With cookieCache adding a second cookie this **broke login entirely**; Set-Cookie now goes through `multiValueHeaders`. Worth knowing about even independently of cookieCache.

**Frontend** — persistence moved from the (silently-failing) synchronous localStorage persister to IndexedDB (async, no quota cliff). The existing in-memory `fetchedMap` mount-gate is kept, so a hard refresh still revalidates collaborative data (scores, lists, awards, comments): cached data paints instantly, a background refetch runs on the first mount(s) of the session, and repeat navigation stays quiet. `useQueueReview` optimistically removes the item from its source list; mutation invalidations run with `Promise.all`.

## Not in this PR (requires Netlify console)

**Move the functions region to us-west-2** (Site settings → Functions). This is the single biggest lever — it turns every remaining ~65-70ms query hop into ~1-5ms — but it's an account setting, not code.

## Verification

- `type-check`, `lint`, all 252 tests pass.
- Driven end-to-end via `netlify dev` + headless Chrome: login, reviews gallery, detail drawer (cast loads on demand, "See all (138)" for The Dark Knight), statistics leaderboards (Most Watched Actors populated from `/reviews/cast`), lists page, add-movie (Paddington), move-to-reviews (optimistic removal + appears in reviews).
- Note: one leftover test row on shared dev — **Paddington (2014) added to cole-test and moved to its reviews list** (left in place per no-destructive-ops policy on shared dev).


netlify/functions/repositories/ClubRepository.ts
```diff
@@ -106,14 +106,23 @@ class ClubRepository {
   }
 
   async isUserInClub(clubId: string, userId: string, isLegacy?: boolean) {
```
cole-adams: Does anywhere use the legacy column anymore? Can we just remove this parameter? And maybe log a follow-up to drop the column

cole-adams (changes requested):

src/main.ts
```diff
@@ -28,33 +28,32 @@ import MenuCard from "@/features/clubs/components/MenuCard.vue";
...
-            return false;
-          } else {
-            fetchedMap.set(query.queryHash, (refetchTimes ?? 0) + 1);
-            return true;
-          }
-        },
+        // Serve cached data without a network refetch for a minute; mutations
+        // invalidate their keys explicitly, so navigating between pages stays
+        // instant instead of re-firing every query on each mount.
+        staleTime: 1000 * 60,
```
cole-adams: This makes it so a refresh won't necessarily trigger a new query. For the refresh to see other's scores, this might not work.

I know we have polling for that, but it is also natural for a user to hit the refresh button which would also kill the polling

netlify/functions/repositories/ClubRepository.ts
brian-norman: Good catch — done in 1642e26. The only caller (`auth.ts`) never passed `isLegacy`, so that whole `legacy_id` join branch was unreachable. Removed the parameter, the branch, and the now-unused `isTrue` import; membership is just the single `club_member` index lookup now.

On dropping the column: it is effectively write-only dead weight already. Its only readers (`getLegacyIdForId`, `mapLegacyIds`) have zero callers, and it is only ever written at `club/index.ts:105` with a *random* value. Logged a follow-up to drop it: #427.

src/main.ts
brian-norman: Fixed in 1642e26. You are right that the global 60s `staleTime` + the new IndexedDB persister means a hard refresh rehydrates cached scores instead of starting cold, so within the window it would not re-fetch to show others' scores — and `startScorePoll` only covers the moment right after *you* submit.

I overrode just the reviews-list query (`useReviewsList`) to `staleTime: 0`, so it always revalidates on mount (including a hard refresh): cached data still paints instantly, but a background re-fetch runs so a refresh always reflects everyone's latest scores. The global 60s stays for all other queries, so page-to-page navigation keeps the speedup.

cole-adams: I'm still not completely sold on this. There's a few other things that are kind of "collaborative" in the app. Things like adding to lists, leaving comments, updating in the awards flow. For all of these, a user will be required to wait a full minute, with possibly multiple refreshes to get the fresh data. What's the benefit of this over the old solution? The old solution also made sure we didn't refresh when switching between pages

brian-norman: Good push — you're right, and I'd only patched `useReviewsList`; awards, lists, and comments would all still have gone stale for up to a minute across refreshes.

The root issue was that I conflated two independent changes. The IndexedDB persister is the actual cache win (the old localStorage one was silently blowing the ~5MB quota and starting every session cold). The `staleTime: 60s` was a *separate* change, and it's the part that broke collaboration — because with the persister now rehydrating the cache on load, a fixed staleTime means a hard refresh no longer revalidates. And a time window fundamentally can't distinguish "remounted during the session" from "reloaded the page," which is exactly the distinction the old `fetchedMap` made by living in memory and resetting on load.

So I've reverted just the staleTime part and restored the `fetchedMap` mount-gate as the global default, while keeping the persister. That gives us stale-while-revalidate on every load: cached data paints instantly, a background refetch runs on the first mount(s) of the session so a refresh always reflects everyone's latest, and repeat page navigation stays quiet. It applies to *every* query automatically, so there's nothing to remember to annotate per-surface. Dropped the special-case `staleTime: 0` on the reviews list since the global gate now covers it.

Done in 7a57989.

cole-adams (approved):

PR #425, opened 2026-07-19
Restrict review spotlight actor facts to major or recognizable actors

brian-norman:
## Problem

The review spotlight "Familiar face" fact (`actorMilestone` in `src/features/reviews/reviewFacts.ts`) counted **every** actor in a movie's cast when looking for a recurring-actor milestone. A bit-part player buried deep in the credits could trigger the fact once they'd incidentally appeared in 5+ club movies — not an interesting or recognizable "familiar face."

## Change

An actor now counts as a **major presence** in a film if **either**:

1. **They're top-billed** — in the leading `MAJOR_CAST_SIZE` (5) slice. TMDB returns the cast in billing order and the movie provider preserves it (`orderBy("cast_order")`), so the leading slice is the top-billed cast.
2. **They're a recognizable star** — their personal TMDB popularity clears a `STAR_POPULARITY` bar, independent of this film's billing depth.

The second rule addresses review feedback: a fixed billing cutoff alone drops a genuinely famous actor billed just outside it in a large ensemble (e.g. Tom Holland, billed 12th in *Civil War*). `actorMilestone` only considers the target's major actors as candidates, **and** only counts prior movies where that actor was also major — so the fact tracks a genuinely recurring prominent presence, not incidental cameos. Non-movie works still short-circuit (`majorActorsOf` returns `[]`, gated by `hasElements`).

### Storing the importance signal

TMDB exposes a per-actor `popularity`, but we weren't storing it (`movie_actors` only had `cast_order`). This PR adds it end-to-end:

- **Migration** `20260721_AddActorPopularity.ts` — new `movie_actors.popularity` column, backfilled from TMDB (mirrors the `20260315_AddPersonProfilePaths` batched backfill). Validated on a freshly spawned DB: **613 movies, 0 errors**.
- Threaded through the writer (`movieDetailsUpdater`), both read paths (`movieProvider` aggregation + `tmdb.ts` awards path), the shared types (`TMDBCastMember`, `DetailedMovieData.actors`), and regenerated `db.ts`.

### Calibrating `STAR_POPULARITY` against real data

The threshold was first guessed at `20` against an assumed 0–100 fame scale. Measuring the actual per-actor popularity across the club's ~630 movies showed that was ~3× too high — TMDB's credit popularity is a compressed, daily-trending score:

- Median billed **lead** popularity ≈ **2.1**; only ~1.5% of leads clear 10, and **0%** clear 20 — so `>= 20` promoted nobody (dead rule).
- Recognizable actors billed **outside** the top 5 cluster from ~6 up: Zendaya ~13.8 (billed 6th), Tom Holland ~13.1 (billed 12th), Scarlett Johansson ~11.5, Brad Pitt ~9.3, RDJ ~7.9, Morgan Freeman / Nicole Kidman / Florence Pugh ~6.1. Ordinary supporting cast caps around a ~4 95th percentile.

`STAR_POPULARITY` is therefore set to **6** — the knee where recognizable names give way to background cast, with margin below the star cluster since the score drifts over time.

## Tests

- Existing "counts an actor milestone" (lead) and "ignores a bit-part actor buried deep in the cast" tests still pass.
- Added "counts a recognizable star billed just outside the top-billed slice" — an actor billed 6th with high popularity surfaces the milestone (uses Tom Holland's real ~13 popularity).
- All 24 `reviewFacts.spec.ts` tests pass; full suite 254/254; `type-check` and `lint` clean.

### DIFF
--- a/migrations/schema/20260721_AddActorPopularity.ts
+++ b/migrations/schema/20260721_AddActorPopularity.ts
@@ -1,0 +1,127 @@
+import axios from "axios";
+import { Kysely, sql } from "kysely";
+
+interface TMDBCastMember {
+  id: number;
+  order: number;
+  popularity: number;
+}
+
+interface TMDBCreditsResponse {
+  id: number;
+  cast: TMDBCastMember[];
+}
+
+const BATCH_SIZE = 40;
+const BATCH_DELAY_MS = 1000;
+
+async function fetchCredits(externalId: string): Promise<TMDBCreditsResponse> {
+  const tmdbApiKey = process.env.TMDB_API_KEY;
+  const response = await axios.get<TMDBCreditsResponse>(
+    `https://api.themoviedb.org/3/movie/${externalId}/credits?api_key=${tmdbApiKey}`,
+  );
+  return response.data;
+}
+
+interface MovieRow {
+  external_id: string;
+  title: string | null;
+}
+
+/**
+ * Adds a per-actor `popularity` column to `movie_actors` and backfills it from
+ * TMDB. The reviews-spotlight "Familiar face" fact uses it as a recognizability
+ * signal so a genuinely famous actor billed just outside a large ensemble's
+ * top-billed slice (e.g. Tom Holland in an Avengers movie) still counts as a
+ * major presence. Mirrors the profile_path backfill in
+ * 20260315_AddPersonProfilePaths.ts.
+ */
+export async function up(db: Kysely<unknown>) {
+  await sql`ALTER TABLE movie_actors ADD COLUMN IF NOT EXISTS popularity NUMERIC`.execute(db);
+
+  const { rows: movies } = await sql<MovieRow>`
+      SELECT DISTINCT md.external_id, md.title
+      FROM movie_details md
+      INNER JOIN movie_actors ma ON ma.external_id = md.external_id
+      WHERE ma.popularity IS NULL
+    `.execute(db);
+
+  console.log(`Found ${movies.length} movies needing popularity backfill`);
+  let processed = 0;
+  let errors = 0;
+
+  for (let i = 0; i < movies.length; i += BATCH_SIZE) {
+    const batch = movies.slice(i, i + BATCH_SIZE);
+
+    // Fetch all credits in parallel (TMDB allows 40 req/10s)
+    const results = await Promise.allSettled(
+      batch.map((movie) =>
+        fetchCredits(movie.external_id).then((credits) => ({
+          movie,
+          credits,
+        })),
+      ),
+    );
+
+    const actorUpdates: {
+      externalId: string;
+      actorId: number;
+      popularity: number;
+    }[] = [];
+
+    for (const result of results) {
+      if (result.status === "rejected") {
+        const message =
+          result.reason instanceof Error ? result.reason.message : String(result.reason);
+        console.error(`Error fetching credits: ${message}`);
+        errors++;
+        continue;
+      }
+
+      const { movie, credits } = result.value;
+
+      for (const castMember of credits.cast) {
+        actorUpdates.push({
+          externalId: movie.external_id,
+          actorId: castMember.id,
+          popularity: castMember.popularity,
+        });
+      }
+
+      processed++;
+      console.log(`Processed: ${movie.title ?? movie.external_id}`);
+    }
+
+    // Bulk UPDATE actor popularity
+    if (actorUpdates.length > 0) {
+      const values = sql.join(
+        actorUpdates.map(
+          (u) => sql`(${u.externalId}::VARCHAR, ${u.actorId}::INT8, ${u.popularity}::NUMERIC)`,
+        ),
+      );
+      await sql`
+          UPDATE movie_actors AS ma
+          SET popularity = v.popularity
+          FROM (VALUES ${values}) AS v(external_id, actor_id, popularity)
+          WHERE ma.external_id = v.external_id
+            AND ma.actor_id = v.actor_id
+        `.execute(db);
+    }
+
+    // Delay between batches to respect TMDB rate limits
+    if (i + BATCH_SIZE < movies.length) {
+      await new Promise((resolve) => setTimeout(resolve, BATCH_DELAY_MS));
+    }
+
+    console.log(`Progress: ${Math.min(i + BATCH_SIZE, movies.length)}/${movies.length}`);
+  }
+
+  console.log("\n=== Backfill Summary ===");
+  console.log(`Total movies: ${movies.length}`);
+  console.log(`Successfully processed: ${processed}`);
+  console.log(`Errors: ${errors}`);
+}
+
+export async function down(db: Kysely<unknown>) {
+  await sql`ALTER TABLE movie_actors DROP COLUMN IF EXISTS popularity`.execute(db);
+}
--- a/migrations/schema/20260721_AddActorPopularity.ts
+++ b/migrations/schema/20260721_AddActorPopularity.ts
@@ -51,19 +51,13 @@ export async function up(db: Kysely<unknown>) {
   // for idempotent re-runs (CockroachDB has no transactional DDL, so a mid-run
   // failure leaves the column behind). Keep this one step as raw DDL; the data
   // queries below run through a typed handle.
-  await sql`ALTER TABLE movie_actors ADD COLUMN IF NOT EXISTS popularity NUMERIC`.execute(
-    db,
-  );
+  await sql`ALTER TABLE movie_actors ADD COLUMN IF NOT EXISTS popularity NUMERIC`.execute(db);
 
   const typedDb = db.withTables<MigrationTables>();
 
   const movies = await typedDb
     .selectFrom("movie_details")
-    .innerJoin(
-      "movie_actors",
-      "movie_actors.external_id",
-      "movie_details.external_id",
-    )
+    .innerJoin("movie_actors", "movie_actors.external_id", "movie_details.external_id")
     .where("movie_actors.popularity", "is", null)
     .select(["movie_details.external_id", "movie_details.title"])
     .distinct()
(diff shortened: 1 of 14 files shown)


src/features/reviews/reviewFacts.ts
```diff
@@ -351,16 +351,25 @@ const authorRecord: FactGenerator = (ctx) => {
 
 const isActorMilestone = (n: number): boolean => n >= 5 && n % 5 === 0;
 
+// A "familiar face" milestone should celebrate a recognizable recurring actor,
+// not a bit-part player buried deep in the credits — those aren't interesting
+// to users. TMDB returns the cast in billing order (lead first) and the movie
+// provider preserves it (`orderBy("cast_order")`), so the leading slice is the
+// movie's major, top-billed actors.
+const MAJOR_CAST_SIZE = 5;
```
cole-adams: I'm a little concerned this will exclude important cast members for larger cast sizes. Like I don't think I'd want Tom Holland excluded from an avengers movie because there's 5 people before him. Is there some other importance factor that we can get from TMDB and store?

brian-norman: Good call — you're right that a fixed billing cutoff drops a recognizable actor billed just behind a big ensemble. TMDB does expose a per-actor `popularity` (a global recognizability score, independent of billing position), but we weren't storing it — `movie_actors` only had `cast_order`.

So I did what you suggested: added a `movie_actors.popularity` column (`20260721_AddActorPopularity.ts`, backfilled from TMDB — 613 movies, 0 errors on a spawned DB), threaded it through the movie provider, and changed the fact to treat an actor as "major" if they're **top-billed OR** their personal popularity clears a star bar (`STAR_POPULARITY = 20`). That keeps buried bit-parts out while letting a Tom-Holland-behind-five-Avengers still count. Added a test for exactly that case; all 254 tests + type-check + lint pass.

The `20` threshold is the one fuzzy knob (TMDB's popularity scale is unbounded and time-varying) — easy to tune if it feels off in practice.

brian-norman: Followup: I pulled the actual per-actor `popularity` TMDB returns for all ~630 movies in the club to sanity-check that `20` bar, and it was way too high — turns out TMDB's credit `popularity` is a compressed, daily-trending score, not a 0–100 fame scale.

Measured across the catalog:
- Median billed **lead** popularity ≈ **2.1**; only ~1.5% of leads clear 10, and **0%** clear 20. So `>= 20` promoted literally nobody — the rule was dead.
- The recognizable actors billed **outside** the top 5 cluster from ~6 up: Zendaya ~13.8 (billed 6th in Homecoming), **Tom Holland ~13.1 (billed 12th in Civil War** — the exact case you raised), Scarlett Johansson ~11.5, Brad Pitt ~9.3, RDJ ~7.9, Morgan Freeman / Nicole Kidman / Florence Pugh ~6.1. Ordinary supporting cast caps around a ~4 95th percentile.

So I dropped `STAR_POPULARITY` to **6** — it sits at the knee where recognizable names give way to background cast, and safely below the Holland/Zendaya cluster (the score drifts day to day, so I left margin). The 47 distinct actors it would admit across the whole catalog read as a clean who's-who. Comment now documents the calibration; test uses Holland's real ~13 instead of a fictional value. (`1c543c7`)

migrations/schema/20260721_AddActorPopularity.ts
```diff
@@ -0,0 +1,127 @@
...
+/**
+ * Adds a per-actor `popularity` column to `movie_actors` and backfills it from
+ * TMDB. The reviews-spotlight "Familiar face" fact uses it as a recognizability
+ * signal so a genuinely famous actor billed just outside a large ensemble's
+ * top-billed slice (e.g. Tom Holland in an Avengers movie) still counts as a
+ * major presence. Mirrors the profile_path backfill in
+ * 20260315_AddPersonProfilePaths.ts.
+ */
+export async function up(db: Kysely<unknown>) {
+  await sql`ALTER TABLE movie_actors ADD COLUMN IF NOT EXISTS popularity NUMERIC`.execute(db);
```
cole-adams, line 65: All of these queries should use type-safe kysely queries. If you need, use the `.withTables` util with a snapshot of the current tables you need to make everything type safe

netlify/functions/utils/providers/movieProvider.ts
```diff
@@ -64,6 +65,13 @@ function summaryQuery(externalIds: string[]) {
         .select([
           "external_id",
           sql<string[]>`array_agg(actor_name ORDER BY cast_order)`.as("cast_names"),
+          // Major cast only (top-billed OR a popularity star), filtered in the
+          // DB so the payload still ships names only — see lib/movie/majorCast.
+          sql<
+            string[]
+          >`array_agg(actor_name ORDER BY cast_order) FILTER (WHERE cast_order < ${MAJOR_CAST_SIZE} OR popularity >= ${STAR_POPULARITY})`.as(
```
cole-adams: Please replace this with a type-safe kysely condition

src/features/reviews/tests/reviewFacts.spec.ts
```diff
@@ -13,18 +14,26 @@ const MEMBER_B = "member-b";
 function movieData(
   opts: {
     directors?: string[];
-    actors?: string[];
+    // An actor is either a plain name or, to exercise the popularity path, a
+    // { name, popularity } pair. Billing order = array order.
+    actors?: (string | { name: string; popularity?: number })[];
```
cole-adams: Why do we need both types here? Can we not just use the new type?

cole-adams (changes requested):

migrations/schema/20260721_AddActorPopularity.ts
brian-norman: Done in 508240c. The backfill now runs through a typed `db.withTables<{ movie_details, movie_actors }>()` snapshot — the `SELECT DISTINCT` is a real `innerJoin` + `.distinct()`, and the bulk `UPDATE ... FROM (VALUES ...)` is now one type-safe per-row `updateTable` per actor (the pg pool bounds concurrency; fine for a one-time backfill).

The one bit I left as raw SQL is the `ADD COLUMN IF NOT EXISTS` — Kysely's `addColumn` has no `ifNotExists`, and we want it idempotent because CockroachDB has no transactional DDL (a mid-run failure leaves the column behind). Added a comment saying so.

netlify/functions/utils/providers/movieProvider.ts
brian-norman: Done in 508240c — replaced with Kysely's aggregate builder: `eb.fn.agg("array_agg", ["actor_name"]).orderBy("cast_order").filterWhere(...)`. The column names now go through `fb.ref("cast_order")` / `fb.ref("popularity")` so they're schema-checked. The comparisons themselves stay in a `sql` fragment only because `cast_order` (Int8) and `popularity` (Numeric) select as `string`, so Kysely rejects a numeric operand; the thresholds bind as params. Compiled it to confirm identical output: `... filter(where ("cast_order" < $1 or "popularity" >= $2))` with params `[5, 6]`.

src/features/reviews/tests/reviewFacts.spec.ts
brian-norman: Good call — collapsed to just `{ name: string; popularity?: number }[]` in 508240c. The string branch was only terseness sugar; call sites now pass `{ name: "Tom Hanks" }` etc.

migrations/schema/20260721_AddActorPopularity.ts
```diff
@@ -0,0 +1,158 @@
...
+    // actor_id (the same bulk-positional-update shape as work_list_item's
+    // position rewrite). Kysely can't express UPDATE ... FROM (VALUES ...) with
+    // typed identifiers, and issuing one statement per cast member instead cost
+    // thousands of round trips per batch — enough to blow the Netlify build
+    // time limit. Grouping keeps the identifiers schema-checked while holding
+    // the statement count at one per movie.
+    await Promise.all(
+      [...actorUpdates].map(([externalId, actors]) => {
+        const whenClauses = actors.map(
+          (a) => sql`WHEN ${a.actorId}::INT8 THEN ${a.popularity}::NUMERIC`,
```
cole-adams: kysely has a type safe `case` operator. Use that instead of raw SQL. Its super important we use the type safe operators in all possible cases

cole-adams (changes requested):

migrations/schema/20260721_AddActorPopularity.ts
brian-norman: Done — the CASE branches now go through `eb.case("actor_id").when(...).then(...)` instead of raw `sql` fragments, so both the operand column and the branch values are built by Kysely.

The simple-`case` form compares each `when` against `actor_id`, so the chain is folded in a loop rather than written out literally; the map of pending updates now holds a `NonEmptyArray` so the chain always has a head branch without re-asserting non-emptiness at the UPDATE.

Compiled output:

```sql
update "movie_actors" set "popularity" = case "actor_id" when $1 then 2.5 when $2 then 13.1 end
where "external_id" = $3 and "actor_id" in ($4, $5)
```

The explicit `::INT8` / `::NUMERIC` casts are gone with the raw SQL — Kysely renders the `then` values as numeric literals and the `when` values as parameters compared against `actor_id`, so CockroachDB infers both.

Re-validated on a freshly spawned DB: 613 movies, 0 errors, 33,041 rows filled with 11,223 distinct values (range 0–19.03), and per-movie spot checks show each actor got their own value rather than a single repeated one. `type-check`, `lint`, and the full suite (263) are clean.

The one remaining raw statement is the `ALTER TABLE ... ADD COLUMN IF NOT EXISTS` — `addColumn` cannot express `IF NOT EXISTS`, which we need for idempotent re-runs given CockroachDB's non-transactional DDL.

cole-adams (approved):

PR #431, opened 2026-07-21
Remove reviews table view; default to gallery

brian-norman:
## What

The reviews page had a table ↔ gallery view toggle (persisted in `localStorage`). Gallery was already the default, so this removes the table view entirely along with the toggle and all the code that only existed to serve it.

## Changes

- Delete `TableView.vue` and the toggle UI in the page header
- Drop the `localStorage`-backed `isGalleryView` state and its `onMounted`/`watch`
- Remove the table-only **edit column** (pencil / delete / share) and its supporting state: `editingTable`, `reviewToDelete`, the delete-confirmation modal, `useShare`, and `useIsInClub`. The gallery has its own delete/share/score-entry flows in `WorkDetailsContent`.
- Remove the now-unnecessary `columnVisibility` juggling — every remaining column is visible in the gallery by default.

## Tests

Tests that previously reached the table's inline "Add score" affordance now drive the gallery details drawer instead. Added a `scrollIntoView` mock to the shared test setup, which the drawer's card-click handler needs under jsdom. The obsolete "switch between gallery and table view" test is removed.

All 93 reviews-feature tests pass.


cole-adams (changes requested):
Correct me if I'm wrong, but there was a bunch of extra infrastructure around getting the new score dial and score assist to work for table view. That should be cleaned up as well.

Also, the hover info card can be removed as well

brian-norman:
Cleaned up the table-view leftovers in 5e83c2f:

- **Hover info card** — deleted `MovieTooltip.vue` (it was only rendered by the table's title-column cell).
- **Score dial / Score Assist table wiring** — every remaining `ReviewScore` caller passed `editable: false`, so the inline popover → `ScoreEntryPanel`/Score-Assist path was unreachable. Removed `ReviewScore.vue` entirely (the gallery just needs a read-only score, now inlined into the column cell) and the orphaned `useAnchoredPanel` composable (its only job was escaping the table's overflow clipping). Also dropped the dead `editable` meta plumbing in `GalleryView`/`WorkDetailsContent`.

Net: **+12 / −277**. Score entry still lives entirely in the details drawer (`ScoreEntryDock`/`Modal` → `ScoreEntryPanel` → `ScoreDial`/`ScoreAssistFlow`), unchanged and still covered by the ReviewView tests. type-check, lint, and all 251 tests pass.

cole-adams (approved):

PR #433, opened 2026-07-21
Score Assist: suggest scores to two decimal places

brian-norman:
## Summary

Score Assist previously snapped every suggestion to the nearest 0.5 via a `roundHalf()` helper, even though the score entry dial already accepts finer-grained values (e.g. `8.25`). This changes the suggestion rounding to **two decimal places** so a converged midpoint like `7.25` is preserved rather than pushed to `7.5`.

## Changes

- `src/features/reviews/composables/scoreAssistLogic.ts`
  - Replaced `roundHalf()` (nearest 0.5) with `roundToTwoDecimals()` (`Math.round(score * 100) / 100`).
  - Applied it to all suggestion paths: `matched`, `aboveAll`, `belowAll`, and `converged`. Scores are still clamped to 0–10.
  - Updated the `ScoreAssistResult.suggestedScore` doc comment.
- `src/features/reviews/tests/scoreAssistLogic.spec.ts`
  - Updated the "about the same" case (pivot `7.3` now suggests `7.3`, not `7.5`).
  - Renamed/expanded the midpoint-rounding test to assert two-decimal behaviour: a `7.25` midpoint is kept, and a `7.275` midpoint rounds to `7.28`.

## Testing

- `npx vitest run` on the score assist specs — 36 passed.
- `npm run type-check` — clean.
- `npx eslint` on the changed files — clean.

---
_Generated by [Claude Code](https://claude.ai/code/session_01HKysVKAN7G2tcQZLRKiGTR)_

### DIFF
--- a/src/features/reviews/components/WorkDetailsContent.vue
+++ b/src/features/reviews/components/WorkDetailsContent.vue
@@ -114,11 +114,7 @@
       enter-active-class="transition duration-slow ease-standard"
       enter-from-class="-translate-y-1 opacity-0"
     >
-      <ReviewFactCard
-        v-if="isDefined(reviewFact)"
-        :fact="reviewFact"
-        class="mt-4"
-      />
+      <ReviewFactCard v-if="isDefined(reviewFact)" :fact="reviewFact" class="mt-4" />
     </Transition>
 
     <!-- Synopsis -->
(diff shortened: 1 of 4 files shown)


package-lock.json
cole-adams, line 1: Seems like an unnecesary package-lock changes

cole-adams (changes requested):

package-lock.json
brian-norman: Agreed — the entire 262-line diff was just `dev`/`devOptional` flag churn (no version or integrity changes), so it was pure noise from a stray install. Reverted to main's lockfile and rebased onto main; the PR is now just the two source files.

cole-adams (approved):

PR #436, opened 2026-07-24
Show a toast when adding a work to a watchlist

brian-norman:
## Problem

Adding a movie (or book) to a list from the "Add to list" modal fired the mutation and immediately closed the modal without any confirmation. On mobile this made it unclear whether the action worked — the newly-added item is appended to the bottom of the list, so you had to scroll all the way down to confirm it landed.

## Change

`AddWorkModal.vue`'s `onSelectWork` now passes per-call callbacks to the `useAddListItem` mutation:

- **Success:** `toast.success('Added "<title>" to the list')`
- **Error:** `toast.error('Failed to add "<title>". Please try again.')`

This gives immediate, visible feedback regardless of scroll position, and surfaces failures that were previously silent. Uses the existing `vue-toastification` setup already used elsewhere in the app (e.g. `WatchListView`, `RankingsView`).

## Testing

- `npm run type-check` — passes
- `npm run lint` — passes

https://claude.ai/code/session_0114LwnuDKrQ2T5za3dr3Sdo

---
_Generated by [Claude Code](https://claude.ai/code/session_0114LwnuDKrQ2T5za3dr3Sdo)_

### DIFF
--- a/src/features/watch-list/tests/useAddListItem.spec.ts
+++ b/src/features/watch-list/tests/useAddListItem.spec.ts
@@ -1,0 +1,101 @@
+import { screen, waitFor } from "@testing-library/vue";
+import { http, HttpResponse } from "msw";
+import { vi } from "vitest";
+import { defineComponent, h } from "vue";
+import { globalEventBus, TYPE } from "vue-toastification";
+
+import { WorkType } from "../../../../lib/types/generated/db";
+import { server } from "@/mocks/server";
+import { useAddListItem } from "@/service/useList";
+import { render } from "@/tests/utils";
+
+// Spy on vue-toastification's real global event bus (which the plugin the test
+// harness installs emits to) rather than mocking useToast — the mutation's toast
+// interface is resolved deep inside the service, and asserting the emitted "add"
+// event verifies the real end-to-end path. Toasts are emitted as
+// emit("add", { content, type }); the event name is a private enum, so match it
+// with expect.anything() and assert on the toast payload.
+
+// Minimal host that fires the add mutation on click, mirroring how AddWorkModal
+// calls mutate() and then immediately closes (which unmounts the modal).
+const Harness = defineComponent({
+  setup() {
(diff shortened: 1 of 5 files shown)


src/features/watch-list/tests/useAddListItem.spec.ts
```diff
@@ -0,0 +1,101 @@
...
+    await view.user.click(screen.getByRole("button", { name: "add" }));
+
+    // The Add modal closes (unmounts) the instant the mutation is fired.
+    view.unmount();
+
+    // Only now does the in-flight POST resolve.
+    release();
+
+    await waitFor(() =>
+      expect(emitSpy).toHaveBeenCalledWith(
```
cole-adams: We should never write tests like this. This is testing an implementation detail. The only correct way to test this is to assert that the message appeared on the screen using vue testing library

cole-adams (changes requested):

src/features/watch-list/tests/useAddListItem.spec.ts
brian-norman: Agreed — rewritten to assert the rendered toast text via Testing Library (`3329fad`).

The reason it was on the event bus: VTU stubs `<transition-group>` by default, and vue-toastification renders every toast inside one, so the stub silently drops the toast content and nothing ever reaches the DOM. Fix is a per-file `config.global.stubs = { transition: false, "transition-group": false }`, after which the real toast (with `role="alert"`) renders and `findByText` works.

One wrinkle worth knowing for future toast tests: `render()` installs the toast plugin, and `setup.ts` renders the Pinia helper through `render()` too, so each toast appears in more than one container — a singular `getByText` fails with "found multiple elements". Hence `findAllByText`.

Verified it is not a vacuous pass: stripping `onSuccess`/`onError` from `useAddListItem` fails both tests.

cole-adams (approved):

PR #437, opened 2026-07-25
docs: Rewrite CLAUDE.md and .claude/rules for Claude-5 context engineering

brian-norman:
## Why

Applies the principles from [The new rules of context engineering for Claude 5 generation models](https://claude.com/blog/the-new-rules-of-context-engineering-for-claude-5-generation-models) to `CLAUDE.md` and all five `.claude/rules/` files.

This PR started as a trim of the always-loaded files and grew into a full rewrite, because the audit turned up something worth acting on: **every doc claim that contradicted the code was in a hand-maintained mirror of something the code already stated.** The two problems are the same problem. Nobody updates a prose copy of `ls src/service/`, so the content that costs the most tokens is also the content most likely to be wrong.

Net **442 lines deleted, 122 added**.

## Stale claims found and fixed

Each of these was live in a rules file and contradicted by the working tree:

| File                       | Claimed                                                     | Actual                                                                                                         |
| -------------------------- | ----------------------------------------------------------- | -------------------------------------------------------------------------------------------------------------- |
| `frontend-architecture.md` | route transitions use animate.css (`animate__slideInRight`) | animate.css removed; `<transition name="route">` + `html[data-route-transition]` CSS                           |
| `database.md`              | `club` has `legacy_id`                                      | dropped in a5cf78c                                                                                             |
| `database.md`              | `WorkType` is `(movie)`                                     | `movie` **and** `book`                                                                                         |
| `database.md`              | "Key Tables" list                                           | missing `book_details`, `book_authors`, `book_subjects`, `movie_actors`, `next_work`, `work`, `work_list_item` |
| `frontend-architecture.md` | service layer file list                                     | missing `useDiscussionQuestions.ts`, `useMediaSearch.ts`                                                       |
| `code-quality.md`          | "manual checks that **ESLint** will flag"                   | oxlint (missed in #434)                                                                                        |

The animate.css one was the most costly: it would have sent a reader writing `animate__` classes against a stylesheet that no longer defines them.

## What changed

**Cut — content the repo already states.** The `src/features/` directory tree, "Key Technologies" (prose copy of `package.json`), "Key Frontend/Backend Files", the 40-line REST route inventory, repository/service/table lists, the env-var catalogue, "External Services". Where a reader needs the real answer, the docs now point at the source of truth — `lib/types/generated/db.ts` for schema, the router files for routes.

**Reframed — prohibitions into judgment.** "Never use `as` casts" became _treat an `as` cast as a signal that something upstream is wrong_, with the reasoning attached, so it transfers to cases the rule never anticipated. Same treatment for `watch()` and the type guards. The `lib/checks` section dropped its signature table and "when to use each" mapping in favour of pointing at the file — the signatures are the interface.

**Kept and promoted — what the code can't tell you.** The `icons.ts` silent-triangle trap, the shared-`dev` blast radius, the CockroachDB DDL gotchas, the club-type registry pattern, the `Set-Cookie` header-folding bug. Reading the source never surfaces _"this fails silently by rendering a warning triangle."_ That's what these files are for, and it's most of what's left.

**`testing.md` rewritten against the actual config.** The first draft described a Vitest client/server projects split, strict MSW, and coverage thresholds. `vite.config.ts` has none of that — it's on the unmerged `test-suite-overhaul` branch. The file now documents what `src/tests/setup.ts` really does, including the global `vue-router` mock that pins `route.params.clubSlug` to `"test-club"` for every test, which otherwise reads as magic.

## Correction to this PR's earlier reasoning

The previous description argued that path-scoping `code-quality.md` was safe _because_ `CLAUDE.md`'s always-loaded "Key Conventions" summarized the prohibitions. This rewrite removes that section as duplication — the blog's point about not repeating guidance across locations.

The scoping still holds up, by a different argument: `code-quality.md` globs `src/**`, `netlify/**`, `lib/**`, `migrations/**`, and `scripts/**`, so it loads on essentially any code edit. The conventions are present whenever they could apply; they're just no longer stated twice.

## Follow-up worth noting

That `code-quality.md` glob set makes it effectively always-loaded. At 50 lines of genuinely cross-cutting conventions that's a fine trade, but if it grows, splitting the club-type registry section into a narrower-scoped file is the natural next move.

Docs only; no source or config changes.

### DIFF
--- a/.claude/rules/backend-architecture.md
+++ b/.claude/rules/backend-architecture.md
@@ -27,7 +27,7 @@ A handler is normally middleware + Zod body schema + repository + a helper from 
 
 ## Lists
 
-Lists are arbitrary and user-titled; the legacy `watchlist`/`backlog` enum was removed in `20260407_ArbitraryClubLists`. IDs are UUIDs. A club's _reviews_ list is a system list (`system_type = 'reviews'`) and is deliberately filtered out of the list-collection endpoint, fetched instead through a dedicated reviews-id route and a richer reviews shape. Rename and delete are rejected for system lists.
+Lists are arbitrary and user-titled, keyed by UUID. A club's _reviews_ list is a system list (`system_type = 'reviews'`) and is deliberately filtered out of the list-collection endpoint, fetched instead through a dedicated reviews-id route and a richer reviews shape. Rename and delete are rejected for system lists.
 
 `ListRepository.moveItem` is transactional with `ON CONFLICT DO NOTHING`, so moving into the reviews list and moving between user lists share one code path.
 
--- a/.claude/rules/backend-architecture.md
+++ b/.claude/rules/backend-architecture.md
@@ -126,3 +126,12 @@ Backend uses `loggedIn` (any authenticated user) and `secured` (authenticated + 
 - `netlify/functions/og-image.ts` - Open Graph image generation
 - `netlify/functions/services/SharedReviewService.ts` - Shared review data service
 - `lib/types/*.ts` - Shared type definitions (club, movie, reviews, awards, lists, common, watchlist, etc.)
+
+## Adding a New API Endpoint
+
+1. Create handler in appropriate `netlify/functions/` directory
+2. Use the custom Router class for routing
+3. Add middleware for validation (`validClubSlug`, `validListId` for list-scoped routes) and auth (`loggedIn`, `secured`)
+4. Use Zod schemas for request body validation
+5. Use Kysely with generated types for database queries
+6. Return responses using utility functions from `utils/responses.ts`
--- a/.claude/rules/backend-architecture.md
+++ b/.claude/rules/backend-architecture.md
@@ -5,133 +5,42 @@ paths:
 
 # Backend Architecture (Netlify Functions)
 
-## Custom Router System
+## Custom router
 
-Located in `netlify/functions/utils/router.ts`. A type-safe Express-like router supporting middleware chaining with type transformations, path parameter extraction via `path-parser`, method routing (GET, POST, PUT, DELETE), sub-routers with `use()`, and automatic 404/405 responses.
+`netlify/functions/utils/router.ts` is a hand-rolled, type-safe Express-like router — middleware chaining _with type transformations_, path params via `path-parser`, sub-routers through `use()`, automatic 404/405. Middleware widens the handler's first argument, so a route mounted behind `validClubSlug` receives `clubSlug` already resolved and typed:
 
 ```typescript
-const router = new Router("/api/club");
 router.use("/:clubSlug/list", validClubSlug, listRouter);
-router.get("/:clubSlug", validClubSlug, async ({ clubSlug }, res) => {
-  const club = await ClubRepository.getBySlug(clubSlug);
-  return res(ok(JSON.stringify(result)));
-});
+router.get("/:clubSlug", validClubSlug, async ({ clubSlug }, res) =>
+  res(ok(JSON.stringify(await ClubRepository.getBySlug(clubSlug)))),
+);
 ```
 
-## API Structure
+Routers are mounted from `netlify/functions/club/index.ts` and `member.ts` — read those for the current route surface rather than relying on a list here.
 
-- `/api/og-image` - Open Graph image generation for shared reviews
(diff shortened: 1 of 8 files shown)


.claude/rules/backend-architecture.md
```diff
@@ -5,124 +5,42 @@ paths:
...
+## Request pipeline
+
+A handler is normally middleware + Zod body schema + repository + a helper from `utils/responses.ts` (`ok`, `badRequest`, `unauthorized`, `notFound`, `svg`, `redirect`).
+
+- `loggedIn` — any authenticated user. `secured` — authenticated _and_ a member of the resolved club.
+- `validClubSlug` resolves `:clubSlug`. `validListId` loads `:listId`, asserts it belongs to that club, and exposes `listSystemType` so handlers can gate operations on system lists.
+
+## Lists
+
+Lists are arbitrary and user-titled; the legacy `watchlist`/`backlog` enum was removed in `20260407_ArbitraryClubLists`. IDs are UUIDs. A club's _reviews_ list is a system list (`system_type = 'reviews'`) and is deliberately filtered out of the list-collection endpoint, fetched instead through a dedicated reviews-id route and a richer reviews shape. Rename and delete are rejected for system lists.
```
cole-adams: Why is it relevant to include information on a legacy enum?

brian-norman: Fair — it is not. Removed in 7519bc2; the line is now just "Lists are arbitrary and user-titled, keyed by UUID."

That sentence was inherited from the old file, and it fails this PR's own test: it describes a state the code no longer has, so it can only cost tokens or mislead. Nothing in the current schema hints that lists were once a fixed enum, so there is no confusion for it to preempt — `git log` is the right home for that.

Unrelated but adjacent: `.claude/skills/tanstack-query-vue/SKILL.md:76` still shows a `useList(clubSlug, type: WorkListType.backlog | WorkListType.watchlist)` signature, and `.claude/skills/kysely/SKILL.md:217` uses the dropped enum as a `createType` example. The kysely one is harmless illustration; the tanstack one is a stale API shape. Skills are outside this PR's scope so I left them — happy to take them in a follow-up.

brian-norman: Follow-ups are in this PR now (a078a1b). Chasing the enum turned up more than the two lines I flagged — auditing the skills the way the rules files were audited found several claims contradicted by the code:

**tanstack-query-vue**
- `useList` overload example used `WorkListType`, which no longer exists anywhere in the repo. `useList(clubSlug, listId: MaybeRef<string>)` and `useReviewsList(clubSlug)` are separate functions now — replaced with the real `MaybeRef` + computed-key + `enabled` guard pattern.
- Persistence was documented as localStorage; #421 moved it to IndexedDB via `idb-keyval`. Kept the reason from the `main.ts` comment (the ~5MB quota a large club could silently exceed), since that is the part reading the code does not hand you.
- "Clear `localStorage` between tests" was advice for a persister tests never construct — `src/main.ts` does not run under Vitest.
- `refetchOnMount` was described as "allows first refetch per query hash." It also always refetches an invalidated query, and the counter suppresses after roughly the first two mounts, not the first.
- Optimistic-update examples and the key-convention table now use the exported factories (`listKey`, `reviewsListKey`, `clubListsKey`).

**kysely**
- `createType("work_list_type")` example → `club_type`, which is a real enum.
- The `expressionBuilder` subquery filtered `work_list` on a `type` column dropped in the same migration → rewritten against `system_type`, mirroring `getReviewsListId`.
- Index example was on that same dropped column.

Two judgment calls worth surfacing:

The JSON-column gotcha recommended `result.value as MyType`, which contradicts the `as` rule this PR just rewrote. I reworded it toward Zod and noted `SettingsRepository` still casts — the skill was accurately describing the code, so this is the one spot where a doc and the code now disagree deliberately. Happy to revert that hunk if you would rather the skill keep describing what is there.

I did **not** add the `selectAll()`-shadowing or migrations-already-in-a-transaction gotchas to the kysely skill; `database.md` covers both, and duplicating them is the thing this PR is arguing against.

The diff is larger than the content changes — the format hook reformatted markdown tables and code fences in both files, which had never been through it.

cole-adams (approved):

PR #442, opened 2026-07-29
Cut Netlify build minutes further: timeouts, skip-migrate, preview-DB reuse

brian-norman:
## Did #420 work? Yes — per build. Not yet on the monthly bill.

Measured from Netlify's deploy API, split at #420's merge (2026-07-16 23:25 UTC):

|                                  | Builds | Total     | Avg      | Median   |
| -------------------------------- | ------ | --------- | -------- | -------- |
| Jul 1–16 (old `npm run build`)   | 94     | 233.8 min | 149s     | 146s     |
| Jul 17–29 (slim build)           | 66     | 63.4 min  | **58s**  | **46s**  |

Daily burn: **14.6 → 4.9 min/day, a 67% cut.** #420's "152s → 42s" holds up across 66 real builds.

But the account meter says July used **347 minutes** against the 300 cap — *worse* than June's 315, because half of July still ran the old build. The reconciliation is what pointed at the real problem:

> 297.2 min (successful builds, both eras) + ~50 min (failed builds) ≈ **347** ✓

That also settles a billing question worth knowing: Netlify bills **raw build seconds**, not per-build minutes rounded up. Rounding would have made 160 builds cost ~476 min, not 347. So shaving seconds off every build pays as well as eliminating builds.

## Where the remaining minutes went

**Two wedged deploy previews on 2026-07-24 burned 45 minutes — 13% of the month, from one branch on one day.**

```
2026-07-24  claude/spotlight-fun-facts-major-actors  1102s  "Command did not finish within the time limit"
2026-07-24  claude/spotlight-fun-facts-major-actors  1621s  "Command did not finish within the time limit"
```

18 and 27 minutes, killed by Netlify's timeout, producing nothing.

**Every slow build was a migration-carrying PR.** All ten post-merge builds over 100s were on #422, #428, and #425 — all three ship files in `migrations/schema/`, so all three hit the plugin's CockroachDB `BACKUP`/`RESTORE` path: **~46s median → 124–190s**. The hangs and the slow builds are the same code path.

**Root cause found:** no build-time `pg.Pool` set a connect or query deadline, and `node-postgres` defaults both to `0` — wait forever. On a laptop that's a hang you Ctrl-C; in a metered build it's a socket that bills until the platform's timeout fires.

## What this PR changes

**1. Bound every build-time database call** — new `lib/db/poolOptions.ts`, wired into `schemaMigrator.ts` and `db-spawn.ts` and mirrored inline in the plugin (it runs as plain Node and can't import TS). Plus `execSync` timeouts on subprocesses and `timeout 600` around the build command.

Deliberately **client-side deadlines only** (`connectionTimeoutMillis` + `query_timeout`). The server-side `statement_timeout` travels in the connection startup packet, so a CockroachDB version rejecting it would break *every* build rather than just the wedged one.

**2. Skip `npm run migrate` on non-migration previews** (~90% of builds) — saves a `tsx` cold start, a CockroachDB round trip and a full schema introspection. Those previews point at shared `dev`, which the `onSuccess` hook already advanced to main's schema; and with no migration in the diff, the committed types can't have gone stale, so the staleness check moves inside the migrate branch where it belongs.

The plugin sets `SKIP_SCHEMA_MIGRATE` **only** in that one provably-safe branch. Every other path leaves it unset and migrates as before, so a plugin failure degrades to today's behaviour rather than silently skipping a needed migration.

**3. Reuse an existing `pr_` database when the migration change is purely additive** (~100s saved per qualifying rebuild). The single blob hash becomes a per-migration manifest, so the plugin can tell *which* migrations changed.

If every migration already applied to the preview database still hashes to what was applied, the database is reused and the build's `npm run migrate` applies whatever is new. Anything else — an applied migration whose content changed, a deleted file, a missing cache, an unreadable `kysely_migration` — drops and restores exactly as today.

> **Revised after [Cole's review](https://github.com/Cobresun/movie-club/pull/442#discussion_r3676249407).** This originally also handled the *edited* case, by rolling the migration back with `down()` and letting the build re-apply it. That is removed, because Cole is right and the guard I put around it did not cover the real failure mode:
>
> - The fallback was a `try/catch` → drop-and-restore, which only catches a `down()` that **throws**. It does not catch one that **succeeds while failing to restore state** — the normal case. This repo already ships two: `20260407_ArbitraryClubLists.ts` deletes work lists it can't represent in the old shape ("a destructive rollback by necessity", per its own comment), and `20260104_ConsolidateUserImage.ts` documents losing OAuth images. Both exit 0.
> - A drifted post-`down()` schema can also make the subsequent `npm run migrate` *fail* where it would have succeeded, and the fallback has already run by then. So my "can only save time, never add failures" claim was wrong in both directions.
> - It was also **net slower for the PRs it targeted.** #425 edited `20260721_AddActorPopularity.ts` across three pushes, so the path would have fired three times — and that migration's `up()` backfills 33,042 rows from the TMDB API in ~3min, against the ~100s restore it was replacing.
>
> The additive path keeps the saving without any of this: it never calls `down()` and never re-applies anything. `schemaMigrator down` stays as a manual tool for iterating against a spawned DB, documented as deliberately not automated.

**4. Widen the build-ignore rule** — `.github/`, test/spec files, `*/tests/*`, `*/mocks/*`, lint/format configs, `coverage/`.

## Testing

- **8 tests** for the reuse planner (`npm run test:plugins`, `node:test`, added to CI), covering the additive cases that must reuse and every case that must rebuild. It's pure logic in its own module because Netlify validates a plugin entry point's exports against known event names and rejects anything else — and because vitest is rooted at `src/`, which reaching the plugins would mean widening.
- 263 vitest tests pass; type-check and lint clean.
- `netlify build --dry` parses the config and loads both plugins; the plugin module exports exactly `onPreBuild` / `onSuccess`.
- Exercised for real: `build.sh` in skip mode, and `onPreBuild` across `deploy-preview` / `production` / `branch-deploy` — confirming the skip flag is set in exactly one context and that path needs no database at all.
- Pathspecs verified against the actual file list rather than assumed: `**/tests/` silently missed `src/common/tests/` and `src/features/reviews/tests/`; `*/tests/*` catches them. Final list excludes only test/CI/config files, every build input still included.

## Measured on this PR

Both preview builds on this branch are non-migration previews, so they exercise items 1, 2 and 4:

| Build                          | Time    |
| ------------------------------ | ------- |
| Post-#420 non-migration previews (n=40) | avg **43s**, median 44s |
| This branch, cold cache        | **40s** |
| This branch, warm cache        | **31s** |

~8s off a 43s build — matching the predicted cost of a `tsx` cold start plus a CockroachDB round trip plus schema introspection. Two data points, so treat the exact figure as indicative rather than settled.

Verified from Netlify's own build log that the mechanism works end to end, rather than inferring it from the timing:

```
✓ No schema migrations detected, using dev database
Netlify configuration property "build.environment.SKIP_SCHEMA_MIGRATE" value changed.
```

Item 3 is not exercised here (this branch ships no migrations) — the first migration-carrying PR after this merges will be its real test. Item 1 is a tail-risk guard, not a steady-state saving.

## Estimated build minutes going forward

Current rates, from the 66 successful builds between #420's merge and Jul 28 (12 days), extrapolated to 30 days:

| Build class              | Volume     | Avg   | Now        | After          |
| ------------------------ | ---------- | ----- | ---------- | -------------- |
| Non-migration previews   | ~100/month | 43s   | 71 min     | **~58 min**    |
| Migration previews       | ~25/month  | 143s  | 60 min     | **~48 min**    |
| Production deploys       | ~40/month  | 42s   | 28 min     | 28 min (unchanged — production always migrates) |
| **Total**                |            |       | **158 min** | **~134 min**  |

Roughly a **15% cut on top of #420**, against a 300-minute cap.

Where each number comes from:

- **Non-migration previews**: measured 43s → ~35s above.
- **Migration previews**: the 10 such builds in the window came from 3 PRs (#422, #425, #428). One restore per PR is unavoidable (the first build). Of the remaining 7, only those whose commit touched no migration file now qualify for reuse — on #425 that was 3 of 6 rebuilds. So `10 × 143s` → roughly `7 × 143s + 3 × ~50s`, a ~20% cut. **This is by far the softest of the three estimates**: it rests on one PR's commit pattern, and it is sensitive to how often people edit a migration after pushing it.
- **Ignore rule**: avoids whole builds rather than shortening them, so each one saved is worth its full ~40s. Probably only 2–4 min/month — it depends entirely on how often a push touches nothing but tests or CI config.

An earlier revision of this description claimed ~115 min, on the strength of the rollback-based reuse that has since been removed; ~134 min is the figure to hold this PR to. (And before that, ~90 min, which assumed a lower build volume than the data actually shows.)

**The failure tail is the part that doesn't show up in that table.** July's 347 minutes were ~297 min of successful builds plus ~50 min of failed ones — and 45 of that 50 came from two hung builds on a single day. That tail is now bounded: a wedged build is killed at 10 minutes instead of running 18–27, and the database calls underneath it fail in 30–60s rather than never. On a bad month that guard is worth more than items 2–4 combined — and after the item 3 revision, it is clearly the most valuable thing in this PR.

## Notes for the reviewer

- **The TOML formatter hook de-indents `netlify.toml` wholesale**, so the diff has unrelated churn on the redirects/functions blocks. Restoring the original indentation gets reverted by the hook.
- Unrelated, not fixed here: `database-config.json` / `auth-config.json` aren't gitignored at the repo root (`.gitignore` only covers the `netlify/functions/utils/` paths), so running the plugin locally drops untracked files in the root.
- The `main` type-check breakage noted in an earlier revision of this description is no longer reproducing — `npm run type-check` is clean on this branch.


netlify/plugins/preview-database/index.js
```diff
@@ -289,128 +369,157 @@ const onPreBuild = async ({ utils, inputs }) => {
...
+        let rolledBack = true;
+
+        if (plan.rollback > 0) {
+          console.log(
+            `\n⚠️  ${plan.rollback} already-applied migration(s) from this PR changed since the last build`,
+          );
+          console.log("→ Rolling them back so the build's `npm run migrate` re-applies them\n");
+
+          try {
+            rollbackMigrations(url.toString(), plan.rollback);
```
cole-adams: I'll be honest, I deeply disagree with this change. For a lot of migrations, its very difficult (sometimes impossible) to write a perfect down migration script. This change will rely on down migrations to get the preview DB into its previous state, which will not always be accurate, leading to worse previews.

Second off, the deploy preview is meant to be as close to prod as possible. If we're previewing off a db that went up -> down -> up, we risk the down state being different from how it will be when the migration runs in production

cole-adams (changes requested):

netlify/plugins/preview-database/index.js
brian-norman: You're right, and I've removed it. Both of your points hold, and the guard I put around the rollback did not cover the case that actually matters.

**On down migrations being unreliable.** My defence was the `try/catch` — a failed rollback falls back to drop-and-restore, so "this can only save time, never add failures." That only catches a `down()` that *throws*. It does nothing about one that **succeeds while failing to restore state**, which is the normal way a down migration is wrong. We already ship two:

- `20260407_ArbitraryClubLists.ts` deletes every non-system list beyond the first per club — its own comment calls it "a destructive rollback by necessity"
- `20260104_ConsolidateUserImage.ts` notes "OAuth provider images will be lost when running down migration"

Both exit 0. The plugin would have seen success and previewed against a database quietly missing rows, which is exactly the "worse previews" you described.

**On preview/prod fidelity.** Also correct, and it cuts further than schema drift alone: if `down()` leaves the schema subtly off (constraint names, NOT NULL, enum types), the build's subsequent `npm run migrate` can *fail* where it would have succeeded — and the fallback has already run by that point, so there's no recovery. So "never add failures" was wrong in both directions, not just the one.

**One more thing I found while checking your objection:** it was also net *slower* for the PRs it was aimed at. #425 edited `20260721_AddActorPopularity.ts` across three pushes, so the rollback path would have fired three times — and that migration's `up()` backfills 33,042 rows from the TMDB API in ~3min, against the ~100s restore it was replacing. Backfill migrations are both the most expensive to re-apply and the most likely to get edited, so the optimisation was worst exactly where it was meant to help.

**What's left** (`e701469`): reuse only when the change is purely additive — every applied migration still hashes to what was applied, and the build's `npm run migrate` picks up whatever is new. It never calls `down()`. Anything else drops and restores as before. Editing a migration you've already pushed now costs a full restore, deliberately.

That still covers a real case: on #425, three of six rebuilds touched no migration file at all and each paid a full restore for nothing. `schemaMigrator down` stays as a manual tool for iterating against a spawned DB, documented as deliberately not automated.

I've revised the projection in the description accordingly — ~115 up to ~134 min/month, i.e. a smaller saving than I originally claimed — and marked it as the softest of the three numbers, since it rests on one PR's commit pattern.

---

cole-adams:
I'm struggling to see the value in the added complexity of advance-in-place. I think 95% of cases the developer would be changing an existing migration instead of adding a new one. All this extra complexity seems overboard for the 5% case? When its totally acceptable to grab a fresh DB for those cases?

PR #444, opened 2026-07-30
Add site metrics dashboard at /admin

brian-norman:
## What

First observability surface for the product as a whole — a `/admin` dashboard showing totals, growth, and activity across every club.

- **Totals:** users (+ % verified), clubs (split movie/book), memberships, reviews, comments, works, lists
- **Growth:** new users and clubs per week, reviews logged per week
- **Activity:** engaged users, sign-ins, and active clubs over rolling 7/30-day windows
- **Leaderboard:** busiest clubs by review count

## Why the schema changes are in this PR and not a follow-up

Two things can't be reconstructed later, so every day without them is data permanently lost:

1. **`club` / `club_member` had no timestamps at all**, so "clubs created per week" was unanswerable. Backfilled from each club's earliest review or list item.
2. **Better Auth prunes expired `session` rows**, so historical sign-in counts evaporate from the live tables. `metric_snapshot` is a daily roll-up written by a new `@daily` scheduled function, alongside the two that already exist.

### The nullable `created_at` is deliberate

The first cut used `NOT NULL DEFAULT now()`. Running it produced `newClubs.last7Days: 12` — twelve activity-less test clubs presented as brand new. `NOT NULL DEFAULT now()` conflates *"created today"* with *"we don't know when this was created."*

The column is now added with no default, backfilled from evidence, then given `DEFAULT now()` for future inserts. Clubs with nothing to date them by stay NULL, and SQL's three-valued logic handles the rest — `NULL >= cutoff` is never true, so unknown rows drop out of window counts instead of inflating them. The leaderboard renders `unknown` rather than inventing a date.

Backfilled dates are a **floor** (a club is at least as old as its first review), and the UI says so.

## Authorization

There is no admin concept in the schema — `club_member.role` is a per-club free-form string that no server-side check has ever read. So a new `siteAdmin` middleware checks the session email against **`ADMIN_USER_EMAILS`** (comma-separated, Netlify console).

It **fails closed**: unset or empty authorizes nobody, so a misconfigured deploy hides the dashboard rather than exposing it to every logged-in user. The route itself only checks login; the API is the authority, and a 401 renders an explanatory panel.

> **Deploy note:** `ADMIN_USER_EMAILS` must be set in the Netlify console or `/admin` works for no one.

## Notable implementation details

- `count()` is `int8`, which node-postgres returns as a **string**. Every aggregate goes through one `toCount()` helper — an unconverted count turns `total + total` into `"1212"`, a bug that type-checks and renders fine.
- Metric types are Zod schemas with TS types inferred from them. `metric_snapshot.metrics` is jsonb, so reads come back `unknown` — parsing recovers a typed value without an `as` cast.
- The snapshot **history** schema is deliberately narrower than the full metrics schema. Zod strips unknown keys rather than failing, so old snapshots keep parsing after a new metric is added; a strict schema would silently invalidate accumulated history.
- `GROUP BY date_trunc('week', ...)` omits zero rows, which would compress time on a chart — `fillMissingWeeks` / `mergeWeekly` zero-fill the gaps (unit tested).
- Admin queries are excluded from IndexedDB persistence alongside `user`; site-wide metrics shouldn't outlive the session authorised to see them.

## Refactor included

`WidgetShell.vue` and `chartPalette.ts` move to `src/common/` so the new feature doesn't import from `features/statistics/`, per the dependency direction in `code-quality.md`. Mechanical — 15 import lines.

## Verification

- Migration applied, rolled **down**, and re-applied on a spawned DB. Backfill recovered real dates back to 2020-04-28; 0 rows left at the `now()` default.
- Drove `/admin` in a browser against real data (83 users, 22 clubs, 1,186 reviews). The "Signed in: 1 in 30d" tile counted the test login itself.
- Restarted with `ADMIN_USER_EMAILS` unset: server denied and logged, client showed the "Not available" panel.
- Scheduled function invoked four times → exactly one `metric_snapshot` row (upsert is idempotent).
- 281 tests pass; lint, format, and type-check clean.
- Rebased onto `main` (5 commits). `CLAUDE.md` conflicted — resolved by taking main's rewritten version and folding the `ADMIN_USER_EMAILS` note into the existing console-only-secrets bullet, rather than re-adding the env inventory that rewrite deliberately dropped.
- `main` added `20260721_AddActorPopularity`, which sorts **before** this migration. Verified on a freshly spawned DB that the whole chain applies in order (`…20260720 → 20260721 → 20260729`), and that `npm run codegen` reproduces the merged `lib/types/generated/db.ts` byte-for-byte — so Netlify's stale-types build gate passes.

**Pre-existing on `main`, not from this PR:** `src/tests/utils.ts` has two type-check errors (`RenderOptions` generic arity, implicit `any`). Untouched here — worth a separate fix.


migrations/schema/20260729_AddObservability.ts
```diff
@@ -0,0 +1,85 @@
...
+ *    `awards_temp.data` and `club_settings.value`.
+ */
+export async function up(db: Kysely<unknown>) {
+  await db.schema.alterTable("club").addColumn("created_at", "timestamptz").execute();
+  await db.schema.alterTable("club_member").addColumn("created_at", "timestamptz").execute();
+
+  // Earliest activity per club, across both signals that carry a timestamp.
+  // Clubs with no reviews and no list items match nothing here and stay NULL.
+  await sql`
+    UPDATE club
```
cole-adams: Use type-safe kysely queries. You can use the `.withTables` utility to type the incoming db. Please remember this is your .md files

cole-adams (changes requested):

netlify/functions/repositories/MetricsRepository.ts
```diff
@@ -0,0 +1,754 @@
...
+ * activity, just not toward any user's engagement. Queries that group *by* user
+ * exclude them explicitly.
+ *
+ * `kind` is cast on the first branch only: a UNION takes each column's type from
+ * the leading branch, and an uncast string literal there is `unknown` rather than
+ * `text`, which makes the later `FILTER (WHERE kind = 'review')` comparisons
+ * ambiguous.
+ */
+const ACTIVITY_EVENTS = sql`
+  SELECT review.user_id AS user_id, work_list.club_id AS club_id, review.created_date AS ts,
```
cole-adams: type safe query

migrations/schema/20260729_AddObservability.ts
brian-norman: Done — both backfills now run through `db.withTables<MigrationTables>()`, and the two `SET DEFAULT now()` strings became ``alterColumn(...setDefault(sql`now()`))``. The only raw SQL left is that ``sql`now()`` `` default expression.

Kysely reached further than I expected here: the derived table is a `selectFrom` callback ending in `.as("activity")`, `UNION ALL` is `.unionAll()`, and `UPDATE ... FROM` is `.from()` on the update — so the earliest-activity query stays fully typed.

Verified by compiling both updates (SQL is equivalent to what it replaced) and running the migrator against a spawned throwaway DB: clubs backfilled to their earliest activity, 4 of 26 memberships correctly left NULL, and `information_schema` confirms `now()` defaults on both new columns.

Written down in `.claude/skills/kysely/SKILL.md` (Migrations section + gotcha 4) and `.claude/rules/database.md`, framed as: `Kysely<unknown>` is the reason to use `withTables`, not an excuse for SQL strings — raw `sql` only for DDL the builder can't express.

netlify/functions/repositories/MetricsRepository.ts
brian-norman: Rewritten as typed queries in 761af0c.

`ACTIVITY_EVENTS` is now `activityEvents(qc)`, a builder the four consumers mount with `.with("activity", activityEvents)`, and the queries over it use `count(...).distinct().filterWhere(...)` for the windowed and per-kind counts. Where a SQL function had no builder equivalent (`date_trunc`/`to_char`, `extract(epoch ...)`) the fragment stays, but the column references go in as `eb.ref(...)`/`eb.fn.min(...)` expressions, so the identifiers are still schema-checked. Six hand-written row interfaces are gone — inferred from the queries now.

Two judgement calls worth flagging:

- **The union now leads with the list-item branch.** A UNION takes its column types from the first branch, and `added_by_user_id` is the only nullable `user_id` of the three, so leading with it gives the CTE an honestly nullable column rather than a cast that hides the NULLs from the type system. The `::text` cast on `kind` moved along with it.
- **Club sizes are bucketed in TypeScript** instead of `FILTER (WHERE members BETWEEN ...)`. The bounds would have to compare against a `count()`, whose honest type is the string node-postgres hands back, and typing it as a number to make the comparison compile is the exact bug `toCount` exists to prevent. The database still does the counting.

Verified by differential test rather than by eye: spawned a database, seeded recent activity (the dev snapshot is months stale, so every window, the leaderboard and the median otherwise read zero), and diffed full `getMetrics()` output before and after the rewrite — byte-identical, including the per-kind breakdown and the NULL-attribution list add. Snapshot capture/read-back exercised separately; 298 tests pass.

cole-adams (approved):

PR #455, opened 2026-08-02
Split Vitest into client and server projects; make MSW strict

brian-norman:
> Part 5 of 16 splitting #375 into reviewable pieces. Merge in stack order.

`test.root: "src/"` made netlify/functions/ and lib/ invisible to the
runner, so no backend test could exist. Replace it with two projects:

- client: jsdom + MSW, matching src/**/*.{test,spec}.ts
- server: node, pure units under lib/, netlify/functions/utils/, scripts/

Shared settings sit on the root config and reach both via `extends: true`:
`restoreMocks` so spies cannot leak across tests, and `TZ=UTC` so date
assertions mean the same thing on a developer's machine as on CI.

MSW now runs with `onUnhandledRequest: "error"` — an unhandled request
fails the test instead of silently leaking to the network. This needs the
previously missing `reviews-id` and `reviews/:workId/comments` baseline
handlers, added here.

Also: register the remaining src/main.ts global components in the render
helper so view tests render real markup instead of unresolved custom
elements, disable Vue Query retries in tests so error-path assertions do
not race the 3x exponential backoff, make the IntersectionObserver stub
able to report intersection, and add an ag-charts stub (it throws on mount
in jsdom).

The existing tests pass unchanged.

---

### DIFF
--- a/oxlint.config.ts
+++ b/oxlint.config.ts
@@ -22,17 +22,6 @@ export default defineConfig({
     "@tanstack/query/no-void-query-fn": "error",
     "@tanstack/query/mutation-property-order": "error",
   },
-  overrides: [
-    {
-      // `expect(Repo.method)` and `vi.mocked(Repo.method)` are the idiomatic
-      // way to assert on a mocked module, and they trip unbound-method on
-      // every call — the rule has no signal to give in a test file.
-      files: ["**/*.test.ts", "**/*.spec.ts"],
-      rules: {
-        "typescript/unbound-method": "off",
-      },
-    },
-  ],
   options: {
     typeAware: true,
   },
--- a/oxlint.config.ts
+++ b/oxlint.config.ts
@@ -8,21 +8,11 @@ export default defineConfig({
     "eslint/eqeqeq": "error",
     "eslint/no-var": "error",
     "eslint/prefer-arrow-callback": "error",
-    "eslint/no-restricted-globals": [
-      "error",
-      "event",
-      "length",
-      "stop",
-      "toString",
-      "alert",
-    ],
+    "eslint/no-restricted-globals": ["error", "event", "length", "stop", "toString", "alert"],
     "typescript/method-signature-style": "error",
     "typescript/switch-exhaustiveness-check": "error",
     "typescript/no-non-null-assertion": "error",
-    "typescript/strict-boolean-expressions": [
-      "error",
-      { allowNumber: false },
-    ],
+    "typescript/strict-boolean-expressions": ["error", { allowNumber: false }],
     "import/newline-after-import": "error",
     "@tanstack/query/exhaustive-deps": "error",
     "@tanstack/query/stable-query-client": "error",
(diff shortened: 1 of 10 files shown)


src/mocks/handlers.ts
```diff
@@ -31,6 +31,15 @@ export const handlers = [
   http.get("/api/club/:id/settings", () => {
     return HttpResponse.json({});
   }),
+  http.get("/api/club/:id/list/reviews-id", () => {
+    return HttpResponse.json({ id: "reviews" });
+  }),
+  // The details drawer loads a work's comments as soon as it opens, so any
+  // spec that clicks into a review needs this even when comments aren't its
```
cole-adams: don't need an explanation for why a mock handler exists lmao

brian-norman: Removed.

oxlint.config.ts
```diff
@@ -32,6 +22,17 @@ export default defineConfig({
     "@tanstack/query/no-void-query-fn": "error",
     "@tanstack/query/mutation-property-order": "error",
   },
+  overrides: [
+    {
+      // `expect(Repo.method)` and `vi.mocked(Repo.method)` are the idiomatic
+      // way to assert on a mocked module, and they trip unbound-method on
+      // every call — the rule has no signal to give in a test file.
+      files: ["**/*.test.ts", "**/*.spec.ts"],
```
cole-adams: Based on our discussion, we shouldn't really be mocking anything in our tests with the integration approach. With that in mind, I'm only seeing risk in disabling lint rules for tests

vite.config.ts
```diff
@@ -30,14 +30,50 @@ export default defineConfig({
...
+    // mock state never leaks between tests (a vi.spyOn or mockResolvedValue set
+    // in one test cannot silently change the next). Inherited by both projects
+    // via `extends: true`.
+    restoreMocks: true,
+    // Pin the timezone so date assertions mean the same thing everywhere.
+    // Components format date-only strings with `new Date(...)` (parsed as UTC)
+    // and `toLocaleDateString` (rendered locally), so west of Greenwich a
+    // release date renders as the previous day — CI runs UTC and would
+    // disagree with a developer's machine. Inherited by both projects.
+    env: { TZ: "UTC" },
```
cole-adams, line 37: Will this affect any features where the user sees the current date?

oxlint.config.ts
brian-norman: Agreed — removed the override in fd08113.

It was pre-emptive: it existed for `expect(Repo.method)` style assertions on mocked repositories, which is exactly the thing the integration approach means we won't be writing. Nothing on this branch needed it — `npm run lint` is clean with `typescript/unbound-method` back on, and the 304 tests still pass.

vite.config.ts
brian-norman: No — this is `test.env`, so it's set only in the Vitest process. It isn't part of the app build or runtime config, so nothing a user sees changes.

Worth being straight about the tradeoff, though: it makes our date assertions agree between a laptop and CI, but it also means our tests run in the one timezone where date handling is least likely to break. There is a real user-facing case underneath it — `MovieMetadataGrid.vue` does `new Date(releaseDate).toLocaleDateString(...)` on a date-only string, which parses as UTC midnight and renders locally, so anyone west of Greenwich sees release dates one day early. That bug exists on main today, independent of this PR; the UTC pin just means these tests won't be the thing that catches it.

Happy to fix that separately (either `DateTime.fromISO(...)` like the rest of the app, or parse as local) rather than widen this test-infra PR — say the word and I'll open one.

cole-adams (approved):

PR #456, opened 2026-08-02
Add unit tests for lib/ and the backend utils

brian-norman:
> Part 6 of 16 splitting #375 into reviewable pieces. Merge in stack order.

260 tests across the newly-reachable `server` project: the check/type-guard
helpers in lib/checks, the Google Books client, and the netlify function
utilities — router path matching, response builders, slug generation,
review-score aggregation, TMDB and Google Books providers, the Gemini
client, and email.

The HTTP clients (TMDB, Google Books, Gemini) run real axios against MSW,
matching how the client project tests work — only the network is faked, so
request building, parsing and error handling all execute for real.
`netlify/functions/tests/mocks/handlers.ts` gives every third-party endpoint a
default response (built by `fixtures/external.ts`) and `mocks/server.ts`
registers them as the server project's `setupFiles` with
`onUnhandledRequest: "error"`; a test overrides one response with
`server.use(...)` when it needs something different. Assertions are on what
the functions return, never on the requests they made — the fixtures are
derived from the requested id or query, so a wrong id still fails on the
returned value.

---


netlify/functions/utils/providers/tests/googleBooks.test.ts
```diff
@@ -0,0 +1,125 @@
...
+ * axios is mocked so no real HTTP calls are made. The key is supplied via
+ * vi.stubEnv — never vi.unstubAllEnvs(), which would restore the ambient
+ * environment (unset locally, populated on CI) and make the keyless case flaky.
+ */
+import axios from "axios";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+
+import { getGoogleBooksVolume, searchGoogleBooksVolumes } from "../googleBooks";
+
+vi.mock("axios");
```
cole-adams: Can we use msw to mock api responses instead of mocking axios? The way we do the frontend tests?

netlify/functions/utils/tests/gemini.test.ts
```diff
@@ -0,0 +1,191 @@
...
+ * Tests for netlify/functions/utils/gemini.ts
+ *
+ * axios is mocked so no real HTTP calls are made.
+ * GEMINI_API_KEY is provided/removed via vi.stubEnv.
+ */
+import axios from "axios";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import { z } from "zod";
+
+vi.mock("axios");
```
cole-adams: similar, can we use real axios and mock the http responses?

netlify/functions/utils/tests/movieDetailsUpdater.test.ts
```diff
@@ -0,0 +1,327 @@
...
+      createDriver: () => new DummyDriver(),
+      createIntrospector: (database) => new PostgresIntrospector(database),
+      createQueryCompiler: () => new PostgresQueryCompiler(),
+    },
+    log: (event) => {
+      if (event.level === "query") queries.push(event.query);
+    },
+  });
+
+  const didInsert = (table: keyof DB) =>
```
cole-adams: This DB approach is dumb. Just use test containers like we're planning for the rest of the backend tests. Introspecting the queries to make an assertion is an anti pattern

netlify/functions/utils/tests/slug.test.ts
```diff
@@ -0,0 +1,241 @@
...
+  generateSlugFromName,
+  generateUniqueSlug,
+  isReservedSlug,
+  validateSlug,
+  validateSlugFormat,
+} from "../slug";
+
+// ---------- constants ----------
+
+describe("SLUG_MIN_LENGTH / SLUG_MAX_LENGTH", () => {
```
cole-adams: I don't think we need tests on constants. Literally no point in having to change a test if we decide to change the constants

netlify/functions/utils/tests/slug.test.ts
```diff
@@ -0,0 +1,241 @@
...
+    const slug = "a" + "b".repeat(48) + "a";
+    expect(validateSlugFormat(slug)).toBe(true);
+  });
+});
+
+// ---------- isReservedSlug ----------
+
+describe("isReservedSlug", () => {
+  it("identifies 'new' as reserved", () => {
+    expect(isReservedSlug("new")).toBe(true);
```
cole-adams, line 148: Wondering, is there a way we can enumerate the top level routes in the client router in this test? We do always risk adding new routes and forgetting to add them to the restricted list.

Its also possible that we should change the frontend route to be /club/id/<id> so that we never have to worry about restricted club slugs

netlify/functions/utils/tests/tmdb.test.ts
```diff
@@ -0,0 +1,494 @@
+import axios from "axios";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+
+import { WorkType } from "../../../../lib/types/generated/db";
+import { TMDBConfig, TMDBMovieData } from "../../../../lib/types/movie";
+
+// Vitest hoists vi.mock above all imports, so axios is intercepted before
+// tmdb.ts binds to it regardless of statement order.
+vi.mock("axios");
```
cole-adams: avoid mocking axios?

cole-adams (changes requested):

netlify/functions/utils/providers/tests/googleBooks.test.ts
brian-norman: Done — switched to MSW.

Added an MSW server for the `server` vitest project (`netlify/functions/tests/msw.ts`, wired in as a `setupFiles` entry with `onUnhandledRequest: "error"`), so the backend HTTP clients are tested the same way the frontend ones are: real axios, faked network.

The helper returns the array requests get recorded into, so assertions read the URL that actually went on the wire instead of the arguments handed to a mock:

```ts
const calls = captureGet(VOLUME_URL, () => ({ id: "abc" }));
await getGoogleBooksVolume("abc");
expect(calls[0]?.url.pathname).toBe("/books/v1/volumes/abc");
```

netlify/functions/utils/tests/gemini.test.ts
brian-norman: Done — real axios against MSW handlers now.

`capturePost` records the parsed JSON body, so the request-shape assertions (prompt, `responseSchema`, `temperature`) run against what axios actually serialised rather than the object literal passed to a mock.

Dropping the axios mock also let the `vi.resetModules()` + dynamic-import dance go: `gemini.ts` reads `GEMINI_API_KEY` per call, so a plain static import is enough.

netlify/functions/utils/tests/tmdb.test.ts
brian-norman: Done. This one benefited the most from the switch.

Beyond dropping the mock, keying handlers by endpoint removes the ordered `mockResolvedValueOnce` chains — every `getDetailedMovie` test previously had to queue `/configuration` then `/movie/:id` in exactly the right order. Now `/configuration` is registered once in a `beforeEach` and `/movie/:id` answers per-id, which is also how the parallel-works test picks its response:

```ts
captureGet(MOVIE_URL, (request) => byId[requestedMovieId(request)] ?? null);
```

netlify/functions/utils/tests/movieDetailsUpdater.test.ts
brian-norman: Agreed — deleted `movieDetailsUpdater.test.ts`.

You're right that asserting on compiled SQL tests the query builder rather than the behaviour; the assertions would pass just as happily against a query that's wrong in a way Kysely can't see. Better no coverage here than coverage that gives false confidence, so this is dropped from the PR and belongs with the testcontainers-backed backend tests.

netlify/functions/utils/tests/slug.test.ts
brian-norman: Agreed — removed the `SLUG_MIN_LENGTH` / `SLUG_MAX_LENGTH` / `SLUG_PATTERN` blocks.

The behaviour they were standing in for is already covered through `validateSlugFormat` (min/max boundaries, leading/trailing hyphens, uppercase, spaces, underscores), which is the function callers actually use. Left a comment in place of the deleted blocks so nobody re-adds them.

netlify/functions/utils/tests/slug.test.ts
brian-norman: Looked into this, and I don't think the test is worth adding — the collision it would guard against can't happen.

Club URLs are already namespaced on both ends: the client route is `/club/:clubSlug` (`src/router/index.ts`) and the API is `/api/club/:clubSlug`. So a club with the slug `profile` lives at `/club/profile` and cannot shadow `/profile`. Your second suggestion is effectively the status quo already.

Enumerating top-level routes and asserting each is reserved would actually fail today, and correctly so: `/verify-email`, `/forgot-password`, `/reset-password`, `/newClub`, `/join-club/...` and `/club-not-found` aren't in `RESERVED_SLUGS` and don't need to be. The entries that do earn their place (`api`, `admin`, `share`, `auth`) are about URL namespaces, not individual route records — so a router-derived test would be coupling the list to the wrong thing.

I've documented that reasoning on `RESERVED_SLUGS` instead, so the next person doesn't have to re-derive why the list isn't generated:

```ts
// Club URLs are namespaced (`/club/<slug>` on the frontend, `/api/club/<slug>`
// on the backend), so a club slug cannot actually shadow a top-level route
// today — this list is defensive, kept in case club URLs are ever hoisted to
// the root. It is deliberately not derived from the router: ...
```

Happy to add the enumeration test if club URLs ever move to the root, which is the point at which it would start catching something.

netlify/functions/tests/msw.ts
cole-adams, line 1: What is this? Why are we not following the pattern from the client? Giving each expected URL a mock response, then override the responses in tests as needed. This is weird where we're requiring every caller to give the exact shape, and also capturing and asserting against api calls which is an anti-pattern. Should ONLY ever be asserting against results of these functions not the http requests they make

cole-adams (changes requested):

netlify/functions/tests/msw.ts
brian-norman: Fair — rewritten to follow the client pattern.

`msw.ts` and the capture helpers are gone. `tests/mocks/handlers.ts` +
`tests/mocks/server.ts` now mirror `src/mocks/`: a default response per
expected URL, registered once in `setupFiles`, and a test overrides one with
`server.use(...)` when it needs something different. Payload builders live in
`tests/fixtures/external.ts` (moved down from the integration-harness PR, which
now reuses them instead of adding them).

Every assertion is on what the function returns. The 11 tests that read a
captured URL, query string or request body are deleted.

Request shape is still covered where the fixture can carry it in the response:
`/movie/:id` returns `Movie <id>`, `/volumes/:id` returns `Book <id>`, and the
search handler echoes `q` into the returned volume id — so a wrong id or query
fails on the returned value. The one exception is Google Books' empty-`key`
400, which the handler reproduces, so the provider's deliberate "omit the param
entirely" behaviour is guarded by a test that just asserts the call resolves.

Genuinely dropped: `append_to_response=credits`, `printType`/`orderBy`/
`maxResults`, and Gemini's prompt/`responseSchema`/`temperature` forwarding.
Those are exercised end-to-end by the integration suite in #457.

260 server tests pass (was 271), and coverage of `tmdb.ts`, `gemini.ts` and
`providers/googleBooks.ts` is unchanged, so #466's thresholds still hold.
Stack rebased and pushed.

netlify/functions/utils/providers/tests/googleBooks.test.ts
```diff
@@ -0,0 +1,94 @@
...
+  vi.stubEnv("GOOGLE_BOOKS_API_KEY", "test-books-key");
+});
+
+afterEach(() => {
+  vi.unstubAllEnvs();
+});
+
+describe("getGoogleBooksVolume", () => {
+  it("returns the volume payload for the requested id", async () => {
+    const volume = await getGoogleBooksVolume("zyTCAlFPjgYC");
```
cole-adams, line 34: I suppose its fine but like this low-key seems like a pointless test

cole-adams (approved):

PR #457, opened 2026-08-02
Add the integration test harness and core handler tests

brian-norman:
> Part 7 of 16 splitting #375 into reviewable pieces. Merge in stack order.

Adds an `integration` Vitest project that runs handler-level tests against
a real CockroachDB started by testcontainers, plus the harness it needs:

- setup/globalSetup.ts boots the container and runs the schema migrations
- setup/integration.ts resets shared tables between tests
- setup/externalApis.ts intercepts TMDB/Google Books/Gemini/Resend with MSW
- helpers/http.ts drives handlers through the public API surface, so tests
  assert on status codes and JSON rather than on repository call shapes
- helpers/auth.ts issues real signed session cookies
- helpers/factories.ts builds clubs, members, works and reviews

The project runs single-worker (`fileParallelism: false`) because the
tests share one database, with raised timeouts for container startup and
bcrypt.

First handler suites on top of it: auth, club, member and members. The
remaining handlers follow in the next change.

Requires Docker; `npm run test:unit` skips this project.

---

### DIFF
--- a/netlify/functions/tests/member.test.ts
+++ b/netlify/functions/tests/member.test.ts
@@ -1,0 +1,215 @@
+/**
+ * Integration tests for `netlify/functions/member.ts` — the signed-in user's
+ * own profile: their clubs, display name and avatar.
+ *
+ * The avatar routes run the real Cloudinary SDK against an MSW-intercepted
+ * Cloudinary, so the multipart parsing and the upload/destroy calls are
+ * genuinely exercised. What the profile now looks like is read back off the
+ * club members endpoint, which is where a client sees it.
+ */
+import { http, HttpResponse } from "msw";
+import { describe, expect, it } from "vitest";
+
+import { ClubPreview, Member } from "../../../lib/types/club";
+import { ClubType } from "../../../lib/types/generated/db";
+import { handler as clubHandler } from "../club/index";
+import { handler } from "../member";
+import { signIn, TestSession } from "./helpers/auth";
+import { createClub, SeededClub } from "./helpers/factories";
+import { makeEvent, requester } from "./helpers/http";
+import { requestsTo, server } from "./setup/externalApis";
+
+const api = requester(handler);
+const clubApi = requester(clubHandler);
+
+const AVATAR_URL = "https://res.cloudinary.com/test-cloud/image/upload/avatar.jpg";
+
+/** A minimal multipart/form-data body carrying one file field. */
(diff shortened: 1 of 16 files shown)


netlify/functions/tests/setup/externalApis.ts
```diff
@@ -0,0 +1,117 @@
...
+    HttpResponse.json({ result: "ok" }),
+  ),
+);
+
+server.events.on("request:start", ({ request }) => {
+  externalRequests.push({ method: request.method, url: request.url });
+});
+
+/** Requests made to `host` during this test, most recent last. */
+export function requestsTo(host: string) {
```
cole-adams: We should not be asserting against calls to an external API, including and using this helper is against the integration style pattern. If you need to test that an API doesn't get call again or something, replace the mock with something that throws

netlify/functions/tests/member.test.ts
```diff
@@ -0,0 +1,215 @@
...
+
+  it("returns 401 without a session", async () => {
+    const res = await api.put("/api/member/name", { body: { name: "Anon" } });
+
+    expect(res.statusCode).toBe(401);
+  });
+});
+
+describe("POST /api/member/avatar", () => {
+  it("uploads the file to Cloudinary and shows the returned url on the profile", async () => {
```
cole-adams: Cloudinary is an implementation detail. We just want to test the round trip of the file

cole-adams (changes requested):

netlify/functions/tests/setup/externalApis.ts
brian-norman: Removed the request log and `requestsTo` entirely. Where a test needs an endpoint left alone, it now swaps that handler for one that throws:

```ts
// Adding the work cached the details, so serving them must not need TMDB.
failOnRequest("get", `${TMDB}/movie/550`);
```

Same treatment in list.test.ts and scheduled-work-refresh.test.ts on #458, and the testing rules doc on #466 now says not to assert on outgoing requests.

netlify/functions/tests/member.test.ts
brian-norman: Dropped the Cloudinary assertions from the avatar tests — they now only check the round trip through the profile. Replacing an avatar is covered by overriding the upload handler to return a second url and asserting that url shows up on the profile, and the delete-with-nothing-stored case uses `failOnRequest` rather than counting calls.

cole-adams (approved):

PR #459, opened 2026-08-02
Add tests for the service layer and the auth store

brian-norman:
> Part 9 of 16 splitting #375 into reviewable pieces. Merge in stack order.

Covers the TanStack Query composables in src/service (awards, club,
discussion questions, lists, reviews, TMDB, user) and the Pinia auth
store: query keys, cache invalidation, optimistic updates and their
rollback, and error paths.

---


brian-norman:
Self-review pass applying the feedback from #455/#456/#457 before it lands here: the service specs no longer capture requests and assert on their URL/body.

- Query specs derive the response from the request's params and assert the rendered result — with `onUnhandledRequest: "error"`, a request to the wrong path already fails the test, so the URL assertion added nothing.
- Mutation specs drive a fake that behaves like the API (`listsApi()`, `awardsApi()`, `commentsApi()`) and read the result back through the query a component uses: creating a list makes it appear, deleting a comment takes it off the thread, submitting a score shows on the reviews list.
- "Must not fetch" cases replace a `fetchCalled` flag plus `setTimeout` with a handler that throws, asserting the query stays `loading/idle`.
- Dropped the query-key and `OPTIMISTIC_WORK_ID` assertions — tests on constants only restate the constant.

cole-adams (changes requested):
I honestly don't think we should have tests for these services. With our frontend testing approach, any logic in these hooks should be covered by the tests for the components in which they're used.

Most of these tests are just kinda garbage setting up harnesses and asserting that we made an api call?

cole-adams:
If we do keep these test, they should be just calling the composable and testing the composable, not setting up a bunch of harnesses to run them

brian-norman:
Took both notes. The harnesses are gone and about half the specs went with them.

`withSetup` in `src/tests/utils.ts` runs a composable in an app scope with the same plugins `render` installs (Pinia, Vue Query, Toast) and hands back what it returned, so a spec calls the composable and asserts on its own refs — no `defineComponent`, no string template, no `findByText`:

```ts
const { result } = withSetup(() => ({
  items: useList("test-club", "list-1"),
  remove: useDeleteListItem("test-club", "list-1"),
}));

await vi.waitFor(() => expect(titles(result.items)).toEqual(["Doomed", "Keeper"]));
result.remove.mutate("work-99");
await vi.waitFor(() => expect(titles(result.items)).toEqual(["Keeper"]));
```

On the first point — the ones that were only setting up a harness to watch an API call happen are deleted rather than rewritten: the plain "fetches from this url" queries, `useClubSlug` (it asserted the router mock in `setup.ts`), the name/avatar mutations, and all of `src/stores/tests/auth.test.ts`, whose six cases asserted `createTestingPinia`'s stub defaults. 1966 lines → 495, 48 tests → 25.

What's left is the behaviour the component tests in the rest of the stack don't reach on their own:

- optimistic writes and their rollback (`useUpdateClubSettings`, and the list/comment/awards mutations)
- `useSubmitScore` picking between create and update — both mutations are module-private, so this is the only surface for that branch
- `useAllUserListItems` tagging items with their source list (#421)
- the reactive-key queries (`useList`, `useAwards`) swapping when their ref changes
- the disabled-query guards, which are the tests with teeth: `useDiscussionQuestions` firing on mount costs a real Gemini call.

One thing worth flagging: the old optimistic-settings tests polled for the intermediate value, and the "keeps it on success" case was passing for the wrong reason — `findByRole` matched the optimistic `true` and returned before `onSettled`'s refetch reverted it. Those now record every value `awards` takes through a sync watcher and assert on the sequence, and the success case holds the POST open so the refetch confirms the value instead of the assertion racing it.

Happy to delete the rest too if you'd rather not carry any of it.

---

cole-adams (approved):

PR #460, opened 2026-08-02
Add tests for shared components and composables

brian-norman:
> Part 10 of 16 splitting #375 into reviewable pieces. Merge in stack order.

Covers CommentThread, MovieMetadataGrid, PosterImage, VSideDrawer,
VTable, WorkDescription, and the useShare and watchUntil composables.

The watchUntil suite is what surfaced the temporal-dead-zone crash on the
already-satisfied-predicate path, fixed earlier in this stack.

---


brian-norman:
Self-review pass: CommentThread was picking elements out with `document.querySelectorAll("button[class*='text-gray-500']")` and `querySelectorAll("textarea")[0]`, which break on any restyle. Both now query by accessible name / display value.

src/common/tests/CommentThread.spec.ts
```diff
@@ -0,0 +1,301 @@
...
+    workId: string;
+    userId: string;
+    userName: string;
+    content: string;
+    createdDate: string;
+    spoiler: boolean;
+    userImage?: string;
+  }[],
+) => {
+  server.use(
```
cole-adams: Thinking there belong in a common test helper. Honestly the ones that don't do anything should just be covered in the default handler

src/common/tests/CommentThread.spec.ts
```diff
@@ -0,0 +1,301 @@
...
+      () => new HttpResponse(null, { status: 200 }),
+    ),
+  );
+};
+
+// Wraps CommentThread in a parent component that pre-seeds the auth store with
+// the current-user data before CommentThread's setup() runs.  This is necessary
+// because CommentThread captures `currentUserId` once at setup time via
+// `ref(user.value?.id)`, so the user must be present before the child mounts.
+const AuthWrapperForTest = defineComponent({
```
cole-adams: There's a pattern for this already in `ReviewView.spec.ts` pull out that helper instead of baking your own solution

src/common/tests/CommentThread.spec.ts
```diff
@@ -0,0 +1,301 @@
...
+      },
+    ]);
+
+    render(CommentThread, {
+      props: { workId: WORK_ID, clubSlug: CLUB_SLUG },
+    });
+
+    await screen.findByText("Secret ending revealed");
+
+    const blurredPara = document.querySelector(".blur-sm");
```
cole-adams: This test is proving the feature doesn't work for users with accessibility needs. That's a bug. We should likely find a way to render blur without the actual text being on the screen, then the test should also not use a query selector

src/common/tests/PosterImage.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+    expect(screen.queryByRole("img")).not.toBeInTheDocument();
+  });
+
+  it("does not render an img when imageUrl is an empty string", () => {
+    render(PosterImage, { props: { imageUrl: "" } });
+
+    expect(screen.queryByRole("img")).not.toBeInTheDocument();
+  });
+
+  it("renders a placeholder container regardless of image presence", () => {
```
cole-adams: Whats the point of this test? The user doesn't care if there's a first child div?

src/common/tests/useShare.spec.ts
```diff
@@ -0,0 +1,117 @@
...
+
+const DESKTOP_UA =
+  "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/120.0 Safari/537.36";
+const IPHONE_UA =
+  "Mozilla/5.0 (iPhone; CPU iPhone OS 17_0 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/17.0 Mobile/15E148 Safari/604.1";
+const ANDROID_UA =
+  "Mozilla/5.0 (Linux; Android 14; Pixel 8) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/120.0 Mobile Safari/537.36";
+
+/** jsdom's navigator has no `share`/`clipboard`, so both are installed here. */
+function stubNavigator({
```
cole-adams, line 14: Can we use `@testing-library/user-events` to stub the `clipboard` instead of this? Then the tests should be reading from `navigator.clipboard.readText()` to perform their assertions

src/common/tests/VSideDrawer.spec.ts
```diff
@@ -0,0 +1,56 @@
+import { screen } from "@testing-library/vue";
+
+import VSideDrawer from "../components/VSideDrawer.vue";
+import { render } from "@/tests/utils";
+
+// NOTE: VBackdrop is registered globally in main.ts but the shared test
```
cole-adams: If the `render` method is missing global registrations, just add them, don't stub anything

src/common/tests/VTable.spec.ts
cole-adams, line 1: I actually think this table is completely unused. Please remove it and the test file in this PR or in a separate one

src/common/tests/WorkDescription.spec.ts
```diff
@@ -0,0 +1,38 @@
...
+    });
+
+    expect(screen.getByText("A great film about things.")).toBeInTheDocument();
+  });
+
+  it("renders nothing when overview is empty string", () => {
+    const { container } = render(WorkDescription, { props: { overview: "" } });
+
+    // v-if="overview" hides the wrapper div entirely
+    expect(container.querySelector("p")).not.toBeInTheDocument();
```
cole-adams: What is this testing? We shouldn't be using query selectors

cole-adams (changes requested):

src/common/tests/CommentThread.spec.ts
brian-norman: Done — `logIn(pinia)` is now exported from `src/tests/utils.ts`, and the four reviews specs that each had their own copy import it instead.

The wrapper component only existed because `CommentThread` snapshotted the id (`ref(user.value?.id)`) at setup, so `logIn` after render had no effect. That is a real bug on its own: whenever the session resolves after mount, the author of a comment never gets their edit/delete controls. It is a `computed` now, and the spec signs in after render like everywhere else.

src/common/tests/CommentThread.spec.ts
brian-norman: Agreed, that was a bug — `blur-sm` is a visual filter only, so the spoiler was still read out in full by a screen reader and still sat in the page's text content.

A hidden spoiler is now a `Reveal spoiler` button rendering a masked stand-in (`content.replace(/\S/g, "█")`, `aria-hidden`), so the real text does not enter the DOM until it is revealed. The test follows that button and asserts the content is absent beforehand and present after the click — no query selector.

src/common/tests/PosterImage.spec.ts
brian-norman: Nothing worth testing — removed.

src/common/tests/useShare.spec.ts
brian-norman: Done. `userEvent.setup()` installs the clipboard jsdom lacks, so the tests now assert with `await navigator.clipboard.readText()` and the `writeText` spy is gone. The dismissed-share-sheet case reads back `""` instead of asserting "not called".

`stubNavigator` still installs `navigator.share`, since that one has no user-event equivalent.

src/common/tests/VSideDrawer.spec.ts
brian-norman: It is not missing it — `v-backdrop` is already in `render`'s global components; the stub and that comment were leftovers. Both are gone, and the spec asserts on slot content rather than panel classes.

Kept one note: the `close` emit fires from the slide-out transition's `@after-leave`, which jsdom never reaches, so the Escape test asserts the drawer's content goes away instead.

src/common/tests/VTable.spec.ts
brian-norman: Confirmed — the only references were the `movie-table` global registration in `main.ts` and its mirror in the test render helper. Removed the component, both registrations and the spec in this PR.

src/common/tests/WorkDescription.spec.ts
brian-norman: Removed. That file also had two tests asserting the same thing about jsdom having no layout, which tested nothing.

It now stubs the `scrollHeight`/`clientHeight` pair the overflow check compares, so the Read more / Show less toggle is actually exercised, and the empty-overview case uses `toBeEmptyDOMElement()`.

brian-norman:
Pushed. Two things in here are no longer test-only, flagging them since the PR was titled as a test-only change:

- `CommentThread` no longer blurs the real spoiler text — a hidden spoiler is a `Reveal spoiler` button over a masked stand-in, so the content is out of the DOM (and out of the accessibility tree) until it is revealed.
- `currentUserId` went from `ref(user.value?.id)` to a `computed`. The snapshot meant an author lost their own edit/delete controls whenever the session resolved after the component mounted.

`VTable` and its `movie-table` registration are gone, and the comment-thread specs now drive real send/edit/delete round trips through `commentsApi()` (lifted out of `useReviews.test.ts` into `src/mocks/comments.ts`) rather than one-off `server.use` stubs per mutation.

The rest of the stack (#461–#466) is rebased on this and pushed.

src/common/components/CommentThread.vue
```diff
@@ -83,17 +83,20 @@
...
+               content stays out of the DOM until it is revealed. -->
+          <button
+            v-if="isSpoilerHidden(comment)"
+            type="button"
+            class="w-full cursor-pointer select-none whitespace-pre-wrap text-left text-sm text-gray-200 blur-sm transition-all"
             @click="revealedSpoilers.add(comment.id)"
           >
-            {{ comment.content }}
-          </p>
+            <span aria-hidden="true">{{ maskSpoiler(comment.content) }}</span>
```
cole-adams: Actually this looks super ugly, lets revert this and conditionally render the `aria-hidden`? Then the test would conditionally see the comment?

cole-adams (changes requested):

src/common/components/CommentThread.vue
brian-norman: Reverted — it is the blurred `<p>` again, with `aria-hidden="true"` rendered only on the hidden branch so a screen reader is not read the spoiler it cannot see the blur on. The mask helper is gone. The test now finds the text and asserts it carries `aria-hidden` before the click, and no longer does after.

cole-adams (approved):

PR #461, opened 2026-08-02
Add tests for the auth, clubs, settings and profile features

brian-norman:
> Part 11 of 16 splitting #375 into reviewable pieces. Merge in stack order.

Covers the password reset and email verification flows, the club home
and creation views, MenuCard, the club settings and join-club views,
member invitation, and the profile view.

---


brian-norman:
Self-review pass on the mocking here. The four auth specs replaced `@/lib/auth-client` with `vi.mock`, which only ever confirms the view called the function the test told it to expect.

better-auth captures `fetch` when its module loads, so MSW could not see its requests — resolving `fetch` per call (`customFetchImpl`, one line in `src/lib/auth-client.ts`) fixes that. The specs now serve `/api/auth/*` from MSW like every other API, so the real client builds the request and parses the response, and `src/mocks/handlers.ts` gains the default session handler.

NewClubView asserted on the captured POST body; it now derives the new club's slug from that body and asserts the view navigates there.

.claude/skills/verify/SKILL.md
cole-adams, line 1: What is this? I thought these were unit test PRs?

src/features/clubs/tests/ClubHomeView.spec.ts
```diff
@@ -0,0 +1,100 @@
...
+        await new Promise(() => {
+          /* never resolves */
+        });
+        return HttpResponse.json([]);
+      }),
+    );
+
+    const { container } = render(ClubHomeView);
+    // LoadingSpinner renders a .lds-spinner div
+    expect(container.querySelector(".lds-spinner")).toBeInTheDocument();
```
cole-adams: Shouldn't be using a query selector here. Means the loading is not accessible, BUG

src/features/clubs/tests/ClubHomeView.spec.ts
```diff
@@ -0,0 +1,100 @@
...
+    expect(screen.getByText("user")).toBeInTheDocument();
+  });
+
+  it("renders 4 router-link stubs for core sections (no awards)", async () => {
+    // Reviews, Watchlists, Statistics, ClubSettings
+    const { container } = render(ClubHomeView);
+
+    await screen.findByText("dev");
+
+    const links = container.querySelectorAll("router-link-stub");
```
cole-adams: No query selector, look for each of the router links by text or whatever. Same goes for all the "router-link-stub" tests

src/features/clubs/tests/ClubHomeView.spec.ts
```diff
@@ -0,0 +1,100 @@
...
+  });
+
+  it("opens invite modal when the plus pill is clicked", async () => {
+    const { user, container } = render(ClubHomeView);
+
+    // Wait for members to load
+    await screen.findByText("dev");
+
+    // The invite pill is a div with @click="showInviteModal = true"
+    const plusPill = container.querySelector(".rounded-full.border-2.border-slate-600.bg-gray-500");
```
cole-adams: This is crazy. Should always be using testing library to access elements on the screen. Add aria text if need be

src/features/clubs/tests/MenuCard.spec.ts
```diff
@@ -0,0 +1,58 @@
...
+    expect(screen.getByText("Reviews")).toBeInTheDocument();
+  });
+
+  it("renders the image", () => {
+    const { container } = render(MenuCard, {
+      props: { image: "/test-image.svg" },
+      slots: { default: "Watchlists" },
+    });
+
+    const img = container.querySelector("img");
```
cole-adams: This should be a findByRole

src/features/clubs/tests/MenuCard.spec.ts
```diff
@@ -0,0 +1,58 @@
...
+  });
+
+  it("applies the default primary background color", () => {
+    const { container } = render(MenuCard, {
+      props: { image: "/test-image.svg" },
+      slots: { default: "Statistics" },
+    });
+
+    const button = container.querySelector("button");
+    expect(button?.className).toContain("bg-primary");
```
cole-adams: Not a valuable test. Its okay to change styling, we don't want to have to change the test for that

src/features/clubs/tests/MenuCard.spec.ts
```diff
@@ -0,0 +1,58 @@
...
+  });
+
+  it("applies a custom background color", () => {
+    const { container } = render(MenuCard, {
+      props: { image: "/test-image.svg", bgColor: "blue-500" },
+      slots: { default: "Custom" },
+    });
+
+    const button = container.querySelector("button");
```
cole-adams: Should be find by role

src/features/clubs/tests/NewClubView.spec.ts
```diff
@@ -0,0 +1,99 @@
...
+
+import NewClubView from "../views/NewClubView.vue";
+import { server } from "@/mocks/server";
+import { useAuthStore } from "@/stores/auth";
+import { render } from "@/tests/utils";
+
+const push = vi.fn(() => Promise.resolve());
+vi.mock("vue-router", () => ({
+  useRoute: vi.fn(() => ({ params: { clubSlug: "test-club" } })),
+  useRouter: vi.fn(() => ({ push })),
```
cole-adams: Why do we need to mock the router all the time? Should it be a common util?

src/features/clubs/tests/NewClubView.spec.ts
```diff
@@ -0,0 +1,99 @@
...
+    render(NewClubView);
+
+    expect(screen.getByText("Must be logged in to create a new club!")).toBeInTheDocument();
+  });
+
+  it("shows the club name input when logged in", async () => {
+    const { pinia } = render(NewClubView);
+    const authStore = useAuthStore(pinia);
+    // @ts-expect-error Overwriting readonly computed for test
+    authStore.isLoggedIn = true;
```
cole-adams: We have a util for this already. Use it..!!

src/features/profile/tests/ProfileView.spec.ts
```diff
@@ -0,0 +1,122 @@
...
+import ProfileView from "../views/ProfileView.vue";
+import memberData from "@/mocks/data/member.json";
+import { server } from "@/mocks/server";
+import { useAuthStore } from "@/stores/auth";
+import { render } from "@/tests/utils";
+
+// ProfileView reads the signed-in user from the auth store; the avatar/name
+// mutations call auth.refreshSession() in onSettled, so it must resolve (a
+// testing-pinia action stub returns undefined and .catch() would throw).
+function asCurrentUser(pinia: ReturnType<typeof createTestingPinia>) {
```
cole-adams: We have a util for this. Use it..!!

src/features/settings/tests/JoinClubView.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+    render(JoinClubView);
+
+    expect(screen.getByText("Please log in to join this club")).toBeInTheDocument();
+    expect(screen.getByRole("button", { name: "Log In" })).toBeInTheDocument();
+  });
+
+  it("calls auth.login when 'Log In' is clicked", async () => {
+    const { user, pinia } = render(JoinClubView);
+    const authStore = useAuthStore(pinia);
+    vi.mocked(authStore.login).mockResolvedValue(undefined);
```
cole-adams: The fact that it calls login is an implementation detail. Correct way to test would be that the users avatar shows up and nows there's a logout button

src/features/settings/tests/JoinClubView.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+    await user.click(screen.getByRole("button", { name: "Log In" }));
+
+    expect(authStore.login).toHaveBeenCalled();
+  });
+
+  it("shows the club name and Join button when logged in and invite is valid", async () => {
+    const { pinia } = render(JoinClubView);
+    const authStore = useAuthStore(pinia);
+    // @ts-expect-error Override readonly computed for testing
+    authStore.isLoggedIn = true;
```
cole-adams: Can we avoid overriding like this?

src/lib/auth-client.ts
```diff
@@ -2,6 +2,13 @@ import { createAuthClient } from "better-auth/vue";
 
 export const authClient = createAuthClient({
   // No baseURL needed - auth server runs on same domain at /api/auth
+
+  // better-auth captures `fetch` when the module loads; looking it up per call
+  // instead lets tests intercept auth requests at the network boundary like any
+  // other API call, rather than mocking this module.
+  fetchOptions: {
+    customFetchImpl: (input, init) => globalThis.fetch(input as RequestInfo, init),
```
cole-adams: Don't like the cast here. Why do we need to cast?

login.yaml
cole-adams, line 1: What is this? And how does it help testing

me.png
cole-adams, line 1: Why is this in this PR?

cole-adams (changes requested):

src/features/clubs/tests/ClubHomeView.spec.ts
brian-norman: Fixed, and you were right that it was a bug. `LoadingSpinner` had no role or name at all, so nothing but a class selector could find it. It renders `role="status" aria-label="Loading"` now and the test is `findByRole("status", { name: "Loading" })`.

src/features/clubs/tests/ClubHomeView.spec.ts
brian-norman: Done. The shared render helper stubbed `router-link` with `true`, which is what produced `<router-link-stub>` and dropped the slot content. It now stubs as a plain `<a>`, so links keep their labels and answer to `getByRole("link", { name })`. The two counting tests are gone — one asserts each of Reviews/Lists/Statistics/Club Settings is linked and that Awards is not, the other that Awards appears once the feature is on.

That also removed the local copy of this stub in `AdminDashboardView.spec.ts`.

src/features/clubs/tests/ClubHomeView.spec.ts
brian-norman: Agreed, and same story as the spinner: the invite pill was a `<div>` with a click handler, so it was not focusable, not announced, and not reachable by keyboard. It is a `<button type="button" aria-label="Invite members">` now, and the test clicks `getByRole("button", { name: "Invite members" })`.

src/features/clubs/tests/MenuCard.spec.ts
brian-norman: Done — `findByRole("button", { name: "Reviews" })`, which also covers that the slot text is what names the card.

src/features/clubs/tests/MenuCard.spec.ts
brian-norman: Agreed, both styling tests are deleted.

src/features/clubs/tests/MenuCard.spec.ts
brian-norman: Done, the click test goes through `findByRole("button", { name: "Awards" })`.

src/features/clubs/tests/NewClubView.spec.ts
brian-norman: It already is a common util — `src/tests/setup.ts` mocks `vue-router` for the whole suite — so this local one was redundant and it is gone. It only existed because the shared mock handed out a fresh `push` spy per call and had no `query`, so a spec that wanted to assert navigation or read a token had to bring its own.

Both are fixed at the source: `setup.ts` now holds one router and one route object for the suite, reset before every test. Navigation asserts against `vi.mocked(useRouter()).push`, and a spec that needs a query param mutates `useRoute()`. `VerifyEmailView.spec.ts` had a local mock for exactly that reason and no longer does. Documented in `.claude/rules/testing.md`.

src/features/clubs/tests/NewClubView.spec.ts
brian-norman: You are right, sorry — `logIn(pinia)` from `@/tests/utils`. All three specs use it now (this one, `ProfileView`, `JoinClubView`) and the hand-rolled auth-store setup is gone.

src/features/profile/tests/ProfileView.spec.ts
brian-norman: Same fix — this spec used a local `asCurrentUser` helper, now `logIn(pinia)`. The one thing that helper did which the shared util did not was settle `refreshSession()`, which the profile mutations chain onto in `onSettled`; I moved that into `logIn` rather than keeping a local copy.

src/features/settings/tests/JoinClubView.spec.ts
brian-norman: Agreed, deleted. Clicking "Log In" here just delegates to the auth store, and with testing-pinia stubbing actions there is no visible outcome to assert — the test could only ever restate the implementation. The visible half is already covered by the next test, which I renamed to say so: once logged in, the login prompt is gone and the club name and Join button are on screen.

src/features/settings/tests/JoinClubView.spec.ts
brian-norman: Yes — that was the same problem as the ProfileView comment. All the `@ts-expect-error` overrides in this file are replaced by `logIn(pinia)`.

src/lib/auth-client.ts
brian-norman: Good catch, the cast was not needed — `globalThis.fetch` already accepts what better-auth passes. Removed, and `type-check` is clean without it.

The line itself is worth keeping, and it is the reason four of the specs in this PR are not mock-based. better-auth captures `fetch` when its module loads, so MSW cannot see its requests; resolving `fetch` per call lets the auth views be tested at the network boundary like every other API in the suite, rather than by `vi.mock`-ing `@/lib/auth-client` and asserting the view called the function the test told it to expect.

login.yaml
brian-norman: Nothing — it was a stray local file (a playwright-cli session recording) that should never have been committed. Same for `me.png` and the `.claude/skills/verify` skill. They were dropped in the rebase before you reviewed, so the diff you were reading was one commit behind; none of the three are on the branch now.

me.png
brian-norman: Same as `login.yaml` — a stray local screenshot, already off the branch. Sorry for the noise.

.claude/skills/verify/SKILL.md
brian-norman: Not intentional — it is a local Claude Code skill file that got swept into a commit. Removed along with `login.yaml` and `me.png`; it was already off the branch by the time you reviewed, the diff you saw was a commit behind.

brian-norman:
Pushed as 20e4e17. Every thread is answered inline; the short version:

**Two of the query-selector comments were real bugs, and the fix is in the components.** `LoadingSpinner` had no role or accessible name, and the club-home invite pill was a `<div>` with a click handler — not focusable, not announced, not keyboard-reachable. Those are now `role="status" aria-label="Loading"` and a real `<button aria-label="Invite members">`, and the tests query them by role. `MenuCard`'s image is marked decorative, since the slot text is what names the card.

**`router-link` was the reason `router-link-stub` kept showing up.** The shared render helper stubbed it with `true`, which drops the slot content and leaves nothing to query. It stubs as a plain `<a>` now, so links keep their labels and answer to `getByRole("link", { name })` — that also let the admin dashboard drop its local copy of the same stub.

**On the router mock: it already was a common util, it just was not good enough to use.** The shared mock handed out a fresh `push` spy per call and had no `query`, so any spec that wanted to assert navigation or read a token had to re-mock `vue-router` itself. `setup.ts` now holds one router and one route for the suite, reset before each test, and the two local mocks are gone. Written down in `.claude/rules/testing.md`.

**`logIn(pinia)` everywhere** — NewClubView, ProfileView and JoinClubView. The one thing ProfileView's local helper did that the shared one did not was settle `refreshSession()`, so that moved into `logIn`.

**The stray files were already gone.** `login.yaml`, `me.png` and the `.claude/skills/verify` skill were dropped in a rebase before your review, so the diff you were reading was a commit behind. Sorry for the noise.

Two things surfaced that were not in your review, both rebase damage rather than anything you saw:

- The `customFetchImpl` line in `auth-client.ts` — the one you flagged the cast on — had been **reverted**, taking the four auth specs back to `vi.mock`-ing the auth client. Restored, and the cast is dropped; it was not needed.
- `AddMembers.spec.ts` covered a component main deleted in #476, so it could not even resolve its import. Deleted.

`npm run lint` and `npm run type-check` are clean, and the client suite is 448 passing across 52 files.

cole-adams (approved):

PR #462, opened 2026-08-02
Add tests for the awards feature

brian-norman:
> Part 12 of 16 splitting #375 into reviewable pieces. Merge in stack order.

Covers the year, nominations, rankings, categories and result views, the
AwardRanking and AwardResult components, and AddMovieButton — including
the step transitions that drive which view is shown.

---


brian-norman:
Self-review pass: the three specs that captured a request body now assert what the user gets instead — the year moves to the next step, the reveal shows the winner, the ranking submission confirms with its toast. The ranking fake rejects a voter that isn't a member id, so #397 stays covered without asserting on the call.

src/features/awards/tests/AwardRanking.spec.ts
```diff
@@ -0,0 +1,145 @@
...
+
+  it("re-orders nominations when the right chevron is clicked on the first card", async () => {
+    const { user, emitted } = render(AwardRanking, {
+      props: { award, members, user: currentUser },
+    });
+
+    // For "dev" the initial order is Parasite (10) then Moonlight (20).
+    const chevronRight = screen
+      .getAllByRole("button")
+      .find((btn) => btn.querySelector(".mdi-chevron-right"));
```
cole-adams: Should find the button that has like a "Right" aria label rather than using a style class

src/features/awards/tests/NominationsView.spec.ts
```diff
@@ -0,0 +1,94 @@
...
+          movieData,
+        },
+      ],
+    },
+  ],
+};
+
+const props = { clubAward, clubSlug: "test-club", year: "2024" };
+
+function login(pinia: ReturnType<typeof createTestingPinia>) {
```
cole-adams: We have a util for this... USE IT

src/features/awards/tests/RankingsView.spec.ts
```diff
@@ -0,0 +1,111 @@
...
+          movieData,
+        },
+      ],
+    },
+  ],
+};
+
+const props = { clubAward, clubSlug: "test-club", year: "2024" };
+
+function login(pinia: ReturnType<typeof createTestingPinia>) {
```
cole-adams: We have a util for it... use it

cole-adams (changes requested):

src/features/awards/tests/AwardRanking.spec.ts
brian-norman: Good catch — the class query was papering over a real gap: the chevron buttons had no accessible name at all, so there was nothing to query by. Added `aria-label="Move <title> left|right"` to both, and the specs now use `getByRole("button", { name: "Move Parasite right" })`. Also replaced the `buttons.length >= 3` assertion with one that checks each end of the list only offers the direction it can actually move.

src/features/awards/tests/NominationsView.spec.ts
brian-norman: Done — dropped the local helper for `logIn` from `@/tests/utils`.

src/features/awards/tests/RankingsView.spec.ts
brian-norman: Done — using `logIn` from `@/tests/utils` here too.

cole-adams (approved):

PR #463, opened 2026-08-02
Add tests for the watch-list feature

brian-norman:
> Part 13 of 16 splitting #375 into reviewable pieces. Merge in stack order.

Covers WatchListView, SharedListView, ListItems, AddWorkModal,
ManageListsModal, the list-item details drawer and its content, and the
useRandomPicker composable.

The ListItems suite is what surfaced the EmptyState prop mismatch fixed
earlier in this stack.

---


brian-norman:
Self-review pass: ManageListsModal was asserting on captured POST/PUT/DELETE bodies and URLs. It now runs against a small fake lists API, so creating, renaming and deleting are checked by what the modal shows afterwards. `useAddListItem`'s captured payload assertion went too — the success toast already names the work.

src/features/watch-list/tests/ListItems.spec.ts
```diff
@@ -0,0 +1,105 @@
...
+  });
+
+  it("emits select event when poster card image is clicked", async () => {
+    server.use(nextWorkHandler());
+
+    const rendered = render(ListItems, { props: defaultProps });
+
+    await screen.findByText("The Super Mario Bros. Movie");
+
+    const img = document.querySelector("img");
```
cole-adams: Don't use query selector, query by role and alt text

cole-adams (changes requested):

src/features/watch-list/tests/ListItems.spec.ts
brian-norman: Good catch — the `querySelector` was only there because `WorkPosterCard`'s `` had no `alt`, so there was no accessible name to query by.

Fixed in 451d2d2: the card image now gets `:alt="title"` (matching what `PosterImage` already does), and the test selects it with `screen.getByRole("img", { name: "The Super Mario Bros. Movie" })`. That was the only raw DOM query in the suite. Full frontend suite, lint and type-check pass, and the rest of the stack has been rebased on top.

---

PR #464, opened 2026-08-02
Add tests for the reviews feature

brian-norman:
> Part 14 of 16 splitting #375 into reviewable pieces. Merge in stack order.

Covers GalleryView, the shared review view and its comment thread, the
add-review prompt, and the AI discussion-questions panel. Adds a
reviewTable harness for building review-table fixtures.

---


brian-norman:
Self-review pass: AddReviewPrompt captured the move/add requests with `vi.fn()` and asserted on them. The handlers now behave like the API instead — a move only succeeds for a work actually on that list, and only the reviews list accepts a work found through search — so the prompt closing is what proves it did the right thing.

src/features/reviews/tests/AddReviewPrompt.spec.ts
```diff
@@ -0,0 +1,156 @@
...
+
+  return [
+    http.get("/api/club/:id/list/reviews", () => HttpResponse.json(reviews)),
+    http.post("/api/club/:id/list/:listId/items/:workId/move", ({ params }) => {
+      const moving = listItems.find((item) => item.id === params.workId);
+      if (moving) queue(moving);
+      return new HttpResponse(null, { status: 200 });
+    }),
+    http.post("/api/club/:id/list/:listId/items", async ({ request }) => {
+      queue((await request.json()) as Record<string, unknown>);
```
cole-adams: Don't cast here. Use a zod schema to parse if need be

brian-norman: Good call — the handler now parses with `listInsertDtoSchema`, the same schema `netlify/functions/club/list.ts` uses on this endpoint, so the mock rejects a bad body the way the API would instead of trusting it.

cole-adams (approved):

PR #465, opened 2026-08-02
Add tests for the statistics feature

brian-norman:
> Part 15 of 16 splitting #375 into reviewable pieces. Merge in stack order.

Covers the statistics and insights views, the shared statistics view,
useStatisticsData, every widget (activity, era, genres, people, records,
consensus, outliers, taste similarity, TMDB deviation, score charts, the
leaderboards), member filter chips and the widget shell, plus chart
option and palette construction. Adds a shared statistics fixture module.

ag-charts throws on mount under jsdom, so widget tests render against the
stub added with the Vitest project split.

---


brian-norman:
Heads up — this branch was rebased as part of syncing stack #467 onto `main`, and it needs some follow-up that I didn't want to make up an answer for.

**A conflict I resolved for you.** `main` gained `src/features/statistics/tests/TasteSimilarityWidget.spec.ts` via #449 (Scope taste similarity to the current user), and this branch adds its own version of the same path — an add/add conflict. I merged the two suites rather than picking a side: #449's four scope tests (the "You" tab, non-member viewers, keeping the mode across a scope switch) plus this branch's eight presentation tests (gap figures, top agreements, biggest disagreements, and the render-nothing guards), with the two fixture sets kept separate. All 12 pass. Worth a look to confirm that's the split you'd have chosen.

**11 tests now fail**, all from this branch drifting behind `main` rather than from the rebase itself:

| File | Failures | Cause |
| --- | --- | --- |
| `statisticsWidgets.spec.ts` | 1 | expects the `era` widget; `main` removed "Through the Years" (#469) and added `subjects`, `authors`, `reviewer-leaderboard`, `taste-similarity`, `member-outliers` |
| `InsightsView.spec.ts` | 1 | same widget-set change |
| `chartOptions.spec.ts` | 4 | decade/era chart options — the era widget is gone |
| `PersonLeaderboard.spec.ts` | 5 | avatar rendering, row expand/collapse, keyboard expand, empty state |

`npx vitest run --project client src/features/statistics` reproduces all of them.

I stopped there deliberately: deciding what the statistics widget tests *should* assert now that the widget set has changed is this PR's call, not a mechanical rebase fix.

brian-norman:
Self-review pass: dropped the `baseChartOptions` / `baseLegendOptions` / `axisLabelFontSize` specs. They only restated constants, so changing a padding value would have meant changing a test. The `memberSeriesColor` tests stay — that one has behaviour (fixed slots, wrap-around).

cole-adams (approved):

PR #466, opened 2026-08-02
Enforce coverage thresholds; document the test suite

brian-norman:
> Part 16 of 16 splitting #375 into reviewable pieces. Merge in stack order.

Coverage now spans src/, lib/ and netlify/functions/ rather than src/
alone — the backend was previously uncounted, which made the reported
number roughly 3x optimistic. Istanbul thresholds (82/70/77/83) sit just
below the levels the suite reaches and fail the build on regression.

Rewrites .claude/rules/testing.md for the three-project layout: what
src/tests/setup.ts does globally, the client gotchas that each cost an
hour (IntersectionObserver, ag-charts under jsdom, userEvent replacing
navigator.clipboard, the statistics fixture thresholds), and the
integration-test contract — arrange and assert through the same
interfaces a client has, mock at the network boundary and nowhere else.

Notes the Docker requirement in CLAUDE.md.

---


brian-norman:
Self-review pass: `.claude/rules/testing.md` now documents the rule the rest of the stack was rewritten to follow — never assert on a request, derive the fake's response from it or let the fake behave like the server, and use a throwing handler for "must not be called". Also notes the auth-client/MSW setup. Coverage thresholds are unchanged.

cole-adams (approved):

brian-norman:
The failing `checks` job here isn't caused by this PR's diff — `main` is red on its own.

`vue-tsc` fails with `TS2307: Cannot find module '@/mocks/agCharts'` across the six statistics spec files. #514 deleted `src/mocks/agCharts.ts` as unused (true when it was written), and #465 merged right after with tests that import it. Neither saw the other.

Fix is up in #516, which restores the mock on `main`. Once that merges, rebase/merge `main` into this branch and the check should go green with no change to the coverage-threshold work.

---

PR #484, opened 2026-08-14
Unify duplicated request-body parsing, num coercion, and formatDate helpers

brian-norman:
## Summary

Scanned the codebase for similar-yet-slightly-divergent abstractions (duplicated utility functions, repeated boilerplate) and consolidated the three genuine, low-risk clusters found. Superficially-similar code that turned out to be intentional divergence (e.g. `WorkPosterCard` vs `WorkSearchCard`, drawer wrappers with only 2 call sites, per-provider cache-miss logic) was deliberately left alone — see "What wasn't touched" below.

### 1. Request-body parse/validate boilerplate → `parseBody()`

21 route handlers across `netlify/functions/club/{index,list,reviews,settings}.ts` and `club/awards/{category,nomination,ranking,step}.ts` repeated the same three-line shape:

```ts
if (!hasValue(event.body)) return res(badRequest("Missing body" | "No body provided"));
const body = someSchema.safeParse(JSON.parse(event.body));
if (!body.success) return res(badRequest("Invalid body" | "Invalid body provided"));
```

This had already drifted (`"Missing body"` vs `"No body provided"`, `"Invalid body"` vs `"Invalid body provided"`) with no functional reason, and had one real behavioral gap: `club/list.ts`'s `/reorder` and `/:listId/reorder` routes wrapped `JSON.parse` in try/catch and returned 400 on malformed JSON; every other of the ~19 other call sites let a malformed-JSON `SyntaxError` propagate into the router's catch-all (`router.ts`), which returns a generic 500 instead of a 400.

Added `netlify/functions/utils/parseBody.ts`:

```ts
export function parseBody<T>(
  event: HandlerEvent,
  schema: ZodType<T, ZodTypeDef, any>,
  res: (data: HandlerResponse) => RouterResponse,
): T | RouterResponse
```

Callers now do:

```ts
const body = parseBody(event, someSchema, res);
if (isRouterResponse(body)) return body;
```

This is the same `isRouterResponse`-composition idiom already used by `secured`/`siteAdmin` in `utils/auth.ts`, so it isn't a new pattern for this codebase. All 21 sites now consistently return 400 on missing body, malformed JSON, *or* a failing schema, with unified messages (`"No body provided"` / `"Invalid body"`).

**Breaking change / migration note:** the exact error message text for a handful of routes changed (`"Missing body"` → `"No body provided"`, `"Invalid body provided"` → `"Invalid body"`). No test or frontend code asserted on these strings (verified via search), but flagging in case any external API consumer does.

### 2. `num()` column-coercion helper → `numOrUndefined()`

`netlify/functions/utils/providers/movieProvider.ts` and `bookProvider.ts` each defined a byte-for-byte identical helper:

```ts
function num(value: string | null): number | undefined {
  return isDefined(value) ? Number(value) : undefined;
}
```

Both files already import `MediaProvider`/`RefreshResult` from the shared `./types` module, so `numOrUndefined` (renamed for clarity now that it's exported) now lives there instead.

### 3. `formatDate()` helper → `src/common/workDisplay.ts`

`WorkDetailsContent.vue` and `ListItemDetailsContent.vue` each defined an identical:

```ts
const formatDate = (dateString: string) => DateTime.fromISO(dateString).toLocaleString(DateTime.DATE_MED);
```

Moved next to the existing `formatRuntime()` helper in `src/common/workDisplay.ts`, which both files already import from.

## What wasn't touched

A couple of other near-duplicates were considered and left alone as judgment calls rather than clear wins (per this repo's "don't introduce abstractions beyond what the task requires" guidance):

- `WorkDetailsDrawer.vue` / `ListItemDetailsDrawer.vue` share a responsive-drawer shell, but only 2 call sites exist today — worth revisiting if either is touched again, not urgent on its own.
- `useAwards.ts` hand-writes its `["awards", clubSlug, year]` query key 7 times instead of using a key-factory function (the convention `useList.ts` follows). Real, but a smaller/lower-confidence win and closer to "add a new pattern" than "unify existing duplication."

## Test plan

- [x] `npm run type-check` — clean
- [x] `npm run lint` (oxlint --deny-warnings) — clean
- [x] `npm run format:check` (oxfmt) — clean
- [x] `npm test` — 570/570 passing (added 6 new tests in `netlify/functions/utils/tests/parseBody.test.ts` covering valid body, missing body, empty body, malformed JSON, failing schema, and `.default()` handling)

---
_Generated by [Claude Code](https://claude.ai/code/session_01KH3MEWh7rByYda7P2WHm8P)_


netlify/functions/club/settings.ts
```diff
@@ -26,15 +26,10 @@ const updateSettingsSchema = z.object({
...
+  if (isRouterResponse(body)) return body;
 
-  const body = updateSettingsSchema.safeParse(JSON.parse(event.body));
-  if (!body.success) return res(badRequest("Invalid body"));
-
-  const settings = await SettingsRepository.updateSettings(
-    clubId,
-    body.data as Partial<ClubSettings>,
-  );
+  const settings = await SettingsRepository.updateSettings(clubId, body as Partial<ClubSettings>);
```
cole-adams: I know this is existing, but can we use a zod schema here instead of casting?

netlify/functions/utils/tests/parseBody.test.ts
```diff
@@ -0,0 +1,69 @@
+import { HandlerEvent } from "@netlify/functions";
+import { describe, expect, it } from "vitest";
+import { z } from "zod";
+
+import { parseBody } from "../parseBody";
+import { createRouterResponse, isRouterResponse } from "../router";
+
+function makeEvent(body: string | null): HandlerEvent {
```
cole-adams: Don't we already have a `makeEvent` helper? You might need to rebase

netlify/functions/utils/tests/parseBody.test.ts
```diff
@@ -0,0 +1,69 @@
...
+      createRouterResponse,
+    );
+    expect(isRouterResponse(result)).toBe(false);
+    expect(result).toEqual({ name: "club" });
+  });
+
+  it("returns a 400 when the body is missing", () => {
+    const result = parseBody(makeEvent(null), schema, createRouterResponse);
+    expect(isRouterResponse(result)).toBe(true);
+    if (!isRouterResponse(result)) throw new Error("expected a RouterResponse");
```
cole-adams: Can we use vitest `assert`? It does the type narrowing so we don't need these if statements

netlify/functions/utils/parseBody.ts
```diff
@@ -0,0 +1,37 @@
...
+ * JSON", and "fails schema" cases with a single 400 response each.
+ *
+ * `Input` is pinned to `any` rather than left to default to `Output`: with it
+ * defaulted, a schema using `.default()` (output field required, input field
+ * optional) makes TS infer `T` as the union of both, silently widening parsed
+ * fields back to optional.
+ */
+export function parseBody<T>(
+  event: HandlerEvent,
+  schema: ZodType<T, ZodTypeDef, any>,
```
cole-adams: cringe any here. Why was this not caught by the `no-explicit-any` lint rule?

cole-adams (changes requested):

netlify/functions/club/settings.ts
brian-norman: Done — moved the schema into `SettingsRepository` and derived the types from it (`ClubSettings = z.infer<typeof clubSettingsSchema>`, plus a `clubSettingsUpdateSchema`/`ClubSettingsUpdate` for the partial patch shape). `updateSettings` now takes `ClubSettingsUpdate`, so the handler passes the parsed body straight through with no cast. Also killed the sibling cast in `getSettings`, which was asserting the shape of the untyped JSON column — it now `safeParse`s and falls back to defaults.

netlify/functions/utils/tests/parseBody.test.ts
brian-norman: Yep — rebased onto main and switched to `makeEvent` from `tests/helpers/http`.

netlify/functions/utils/tests/parseBody.test.ts
brian-norman: Switched to `assert(isRouterResponse(result))` — the narrowing works, so the `if`/`throw` lines and the redundant `expect(isRouterResponse(...)).toBe(true)` are gone.

netlify/functions/utils/parseBody.ts
brian-norman: Gone — the signature is now generic over the schema (`<S extends ZodType>`) and returns `z.output<S> | RouterResponse`, which gets the same `.default()` behaviour without writing `any`. Verified narrowing still produces a real type, not `any`.

As for why lint stayed quiet: `typescript/no-explicit-any` is in oxlint's `restriction` category, which is off by default, and `oxlint.config.ts` only enables the `correctness` defaults plus the rules listed explicitly — `no-explicit-any` is not among them. Happy to add it in a follow-up if you want it enforced repo-wide.

cole-adams (approved):

cole-adams (approved):

PR #487, opened 2026-08-16
Add support for deleting your own review score

brian-norman:
## What

Scoring a work was a one-way door: you could create a score and edit it, but never take it back. The only "delete" in the drawer (`useDeleteReview`) removes the whole work from the club's reviews list — the wrong tool for "I rated the wrong movie", since it destroys everyone's scores.

This adds a per-member score deletion: `DELETE /api/club/:clubSlug/reviews/:reviewId`, owner-only. The work stays on the reviews list with everyone else's scores intact.

## Backend

- **`netlify/functions/club/reviews.ts`** — new `DELETE /:reviewId` behind `secured`. 404 for a review that doesn't exist or belongs to another club, 401 for someone else's review.
- **`ReviewRepository`** — `getById` → `findById`, returning `undefined` instead of throwing so an unknown id answers 404 rather than 500 (the existing `PUT` handler now benefits from this too). Its `selectAll()` is scoped to `selectAll("review")`: `work_list` has an `id` of its own that was shadowing the review's — the exact gotcha `.claude/rules/database.md` calls out. Adds `deleteById`.

## Frontend

- **`useDeleteScore`** (`src/service/useReviews.ts`) — optimistically drops the member's entry from the cached reviews list and recomputes the synthetic `average`, mirroring the server's `buildReviewScores` (including its empty map once nobody has scored). Also retires any score poll left running from an earlier save on the same work, so a late poll response can't paint the removed score back.
- **`ScoreEntryPanel`** — a quiet "Remove my score" action, present only once a score exists, with an inline two-step confirm. Putting it in the panel means both hosts (the desktop `ScoreEntryDock` and the mobile `ScoreEntryModal`) get it unchanged, and the confirm doesn't need to stack a third overlay on top of a drawer/modal. It emits `submit` but never `saved`, so the host closes the panel without firing the just-saved celebration.
- **`WorkDetailsContent`** — clarifies the club-wide "Delete review" confirmation so the two destructive actions can't be mistaken for each other.

## Tests

- `netlify/functions/tests/reviews.test.ts` (new) — own score removed while the work and other members' scores survive; empty score map when the last score goes; 401 for another member's review; 401 for a non-member; 404 across clubs and for an unknown id.
- `src/features/reviews/tests/useDeleteScore.spec.ts` (new) — the optimistic cache rewrite, including the average recomputation.
- `ScoreEntryPanel.spec.ts` — no removal control before a score exists; the confirm sends the DELETE and emits `submit` but not `saved`; dismissing the confirm sends nothing.
- `ReviewView.spec.ts` — end-to-end through the drawer: removing your score leaves the work on the list and flips the CTA back to "Rate this movie".
- Added a default MSW handler for the new endpoint.

`type-check`, `lint`, `format:check` and the `client` + `server` vitest projects all pass (570 tests). **The `integration` project was not run** — it needs Docker for its CockroachDB container, which isn't available in this environment, so `reviews.test.ts` is unverified locally and needs CI to confirm.

---
_Generated by [Claude Code](https://claude.ai/code/session_01CjNBc8bdJ2m6a6XzgWQtss)_


src/features/reviews/tests/useDeleteScore.spec.ts
```diff
@@ -0,0 +1,93 @@
...
+ * the cache rewrite — in particular that the club average is recomputed rather
+ * than left reporting a number that still includes the score just removed.
+ */
+
+const CLUB_SLUG = "test-club";
+
+/** Captured from the harness so the test can read the cache the mutation wrote. */
+let queryClient: QueryClient;
+
+const Harness = defineComponent({
```
cole-adams: As commented in the services test review, we should not be setting up these harnesses to render hooks. That PR introduces a helper to add the things necessary to render these hooks. lets wait for that PR to land, then rebase an re-write this test file

src/features/reviews/tests/useDeleteScore.spec.ts
```diff
@@ -0,0 +1,93 @@
...
+function score(id: string, value: number) {
+  return { id, created_date: "2024-05-28T04:46:37.751Z", score: value };
+}
+
+async function removeScore(seeded: DetailedReviewListItem[]) {
+  const { user, pinia } = render(Harness);
+
+  const authStore = useAuthStore(pinia);
+  // @ts-expect-error Overwriting readonly property for testing purposes
+  authStore.user = { id: memberData.id, email: memberData.email, name: memberData.name };
```
cole-adams: As mentioned in one of the test PRs, we're going to pull this logic out into a common helper and use that instead of duplicating it everywhere. Lets wait for that to land and rewrite this test file

src/service/useReviews.ts
```diff
@@ -204,6 +211,61 @@ export function useSubmitScore(clubSlug: string) {
...
+      id: "average",
+      created_date: new Date().toISOString(),
+      score: total / remaining.length,
+    },
+  };
+}
+
+/**
+ * Removes the current user's own score from a work — the "I rated the wrong
+ * thing" escape hatch. Only the caller's review row goes; the work stays on the
```
cole-adams: its whatever but these "escape hatch" comments are cringe af

cole-adams (changes requested):

src/features/reviews/tests/useDeleteScore.spec.ts
brian-norman: Done — #460/#461 have landed, so this is rebased on latest main and the harness is gone. The spec file is deleted; the coverage now lives in `src/service/tests/useReviews.test.ts` as a `useDeleteScore` block on `withSetup` + `logIn`, matching the other mutation tests there.

src/features/reviews/tests/useDeleteScore.spec.ts
brian-norman: Same — the hand-rolled auth setup is replaced by the shared `logIn(pinia)` helper from `src/tests/utils.ts`.

src/service/useReviews.ts
brian-norman: Fair. Cut the "escape hatch" phrasing (and the other flourishes) — those comments now just say what the code does and why.

